Issue #32, opened 2016-01-07
gmock destructor


joakim-brannstrom:
Generated gmocks are missing the destructor.

PR #36, opened 2016-01-31
Bug: add generation of virtual destructor to gmocks

joakim-brannstrom:
The C++ standard states that if a class with one or more virtual functions
but missing the virtual destructor is deleted it is undefined behavior.


joakim-brannstrom:
Fix for #32

Issue #33, opened 2016-01-15
Missing licenses


mstahlberg:
Missing licenses for Dstep (boost) and Clang (Illinois...)

joakim-brannstrom:
See #42

mstahlberg:
Open source license "University of Illinois/NCSA Open Source License" for libclang is still missing.

PR #44, opened 2016-02-09
add Illinois/NCSA license

joakim-brannstrom:
Fix for #33


mstahlberg:
I don't think that the Illinois/NCSA license file shall be in clang/resources. The files there have the MIT style license in their headers.

Illinois/NCSA should only be in the libclang directory.

joakim-brannstrom:
Fixed.
You are right. I should have checked the header instead of assuming.

PR #250, opened 2016-11-15
mac support

ploq:
Seems to fail one test "test.component.analyzer.type.Should be a type of kind 'func'", unable to test on linux at the moment to see if the same test fails there too.

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -15,7 +15,7 @@ elif [[ -d "/usr/lib64/llvm" ]]; then
 elif [[ -d "/Applications/Xcode.app/Contents/Frameworks" ]]; then
     export LFLAG_CLANG_PATH="-L/Applications/Xcode.app/Contents/Frameworks"
     export LFLAG_CLANG_LIB="clang"
-    export DYLD_LIBRARY_PATH="/Applications/Xcode.app/Contents/Framework"
+    export LFLAG_CLANG_ABS_PATH="/Applications/Xcode.app/Contents/Framework"
 fi
 
 if [[ -n "$LFLAG_CLANG_LIB" ]]; then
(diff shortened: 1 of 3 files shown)


joakim-brannstrom:
All pull requests are automatically ran by Travis-CI (linux).
You changes doesn't seem to affect that so good times man.

joakim-brannstrom:
Hmm, I guess the test code is "linux specific".
Does it fail on line 109?
tip to only run the test specific test in debug mode (-h for more options):
```d
./build/unittest -d "test.component.analyzer.type.Should be a type of kind 'func'"
```

ploq:
The tests fails to find clocale, which seems odd since that library exists in Apples version of Clang.

joakim-brannstrom:
Ahh, cool. I guess ClangContext.makeTranslationUnit need an extra "-I" to find the system installation of standard headers.

But lets go with the version-block instead, for now. I don't have a Mac myself so wouldn't be able to fix future bugs either way.

joakim-brannstrom:
Travis-CI is green. Do you feel it is ready to be merged?

ploq:
autobuild does not work at the moment, it fails to find the environment variable `$DYLB_LIBRARY_PATH`, which makes unable to run dextool. Will see if I can figure something out for that

joakim-brannstrom:
If you are wondering what dub is doing in the background you can tell it to be more verbose.
```d
dub build -c unittest -b unittest --verbose
```

joakim-brannstrom:
or
```sh
readelf --dynamic ./build/dextool
```

One way may be to change the path that is searched for libclang.so (I guess it is this one that is the problem).
library path is set to [.] by -rpath in dub.sdl.
If you changed "-rpath" in dub.sdl to wherever it is in your system it should work.

ploq:
Yes, correct, there is a problem with rpath. Problem is that when dmd is linking it says that ld does not have rpath as a option (it does).

ploq:
If Travis-CI is green then everything should be working!

joakim-brannstrom:
Awesome man. Will do!

PR #259, opened 2016-12-12
Missing a character in config.sh

ploq:
Missing character which made running dextool on mac not working!


joakim-brannstrom:
thank you :)

PR #249, opened 2016-11-14
WIP IP plugin

ploq:
atm just basic changes to test dextool

PR #316, opened 2017-03-21
WIP Dovahkiin fuzzer

ploq:
A new attempt on the fuzzer!

Old PR can be found at #249 and in branch old_master https://github.com/ploq/dextool/tree/old_master


joakim-brannstrom:
Nice work team Dovahkiin. That was fast. I see it follows the new plugin structure :)
I'll check the implementation and give some feedback. High and low.

joakim-brannstrom:
I strongly recommend to fix the errors the CI server is complaining about.
At the moment it is:
```bash
Running custom 'unittest' configuration.
Performing "unittest-cov" build using dmd for x86_64.
dextool:libclang ~master: building configuration "library"...
dextool:clang ~master: building configuration "library"...
dextool:dsrcgen ~master: building configuration "library"...
dextool:cpptooling ~master: building configuration "library"...
dextool:unit-threaded ~master: building configuration "library"...
dextool ~master: building configuration "unittest"...
plugin/fuzz/source/dextool/plugin/fuzz/backend/fuzzvariant.d(16,8): Error: module sutenvironment is in file 'sutenvironment/sutenvironment.d' which cannot be read
import path[0] = source/
import path[1] = source
import path[2] = plugin
import path[3] = clang
import path[4] = libclang
import path[5] = dsrcgen/source/
import path[6] = unit-threaded/source
import path[7] = /home/travis/dlang/dmd-2.071.2/linux/bin64/../../src/phobos
import path[8] = /home/travis/dlang/dmd-2.071.2/linux/bin64/../../src/druntime/import
dmd failed with exit code 1.
```

dub.sdl
```diff
@@ -211,6 +211,20 @@ configuration "plugin_uml_debug" {
...
+    dependency "dextool:cpptooling" version="*"
+
+    targetName "dextool-fuzz"
+    targetType "executable"
+    
+    sourcePaths "plugin/source" "plugin/fuzz"
+    importPaths "source" "plugin/source" "plugin/fuzz/source"
+    excludedSourceFiles "source/application/*" "source/test/*" "source/devtool/*" "source/cpptooling/*"
+}
+
```
joakim-brannstrom: I recommend to also add a "plugin_fuzz_debug" with code contracts etc activated. It makes it a lot easier to debug but at the same time keep the release binary "fast, efficient, not-so-chatty".

plugin/fuzz/source/dextool/plugin/fuzz/backend/fuzzvariant.d
```diff
@@ -0,0 +1,628 @@
...
+/** Parameters used during generation.
+ *
+ * Important aspact that they do NOT change, therefore it is pure.
+ */
+@safe pure interface Parameters {
+    static struct Files {
+        FileName hdr;
+        FileName impl;
+        FileName globals;
+        FileName gmock;
```
joakim-brannstrom: seems to be some unnecessary junk. Is this even used in the code? :)

plugin/fuzz/source/dextool/plugin/fuzz/backend/fuzzvariant.d
```diff
@@ -0,0 +1,628 @@
...
+            o.append(makeHeader(dest, ver, custom_hdr));
+            o.include(incl_fname.baseName);
+            o.sep(2);
+            o.append(code);
+
+            return o;
+        }
+
+	prods.putFile(params.getFiles.hdr, outputHdr(modules.hdr, params.getFiles.hdr, params.getToolVersion, params.getCustomHeader));
+        prods.putFile(params.getFiles.impl, output(modules.impl, params.getFiles.hdr, params.getFiles.impl, params.getToolVersion, params.getCustomHeader));
```
joakim-brannstrom: install dfmt and autoformat the files. Easy piecy, all is pretty.

joakim-brannstrom:
```bash
dub fetch dfmt --version0.5.0-beta4
cd dextool
dub run dfmt -- <some file>
```

I think the above uses the .editorconfig in the root of dextool. otherwise supply it with -c.

ploq:
Yup, well aware of that error. We basically just copied the old code from the generate function, which uses a module that we will rewrite (sutenvironment + ipxmlparser).

plugin/fuzz/source/dextool/plugin/fuzz/backend/fuzzvariant.d
```diff
@@ -0,0 +1,628 @@
...
+    import cpptooling.data.symbol.types : USRType;
+    import cpptooling.generator.func : generateFuncImpl;
+    import cpptooling.generator.gmock : generateGmock;
+    import cpptooling.generator.includes : generateIncludes;
+    import cpptooling.data.representation;
+    import cpptooling.analyzer.type;
+
+    //generateIncludes(ctrl, params, modules.hdr);
+    foreach (incl; params.getIncludes) {
+        modules.hdr.include(cast(string) incl);
```
joakim-brannstrom: I think the cast is redundant. FileName is a struct with implicit cast "to" string being allowed.

joakim-brannstrom: Which is one reason why I migrated from the pattern of using Typedef to:
```d
struct MyType {
   string value;
   alias value this;
}
```

compile time safety but still convenient to use in e.g. string concatenations without a horrendous amount of cast's

plugin/fuzz/source/dextool/plugin/fuzz/backend/fuzzvariant.d
```diff
@@ -0,0 +1,628 @@
...
+    }
+    
+    modules.hdr.include("testingenvironment.hpp");
+    modules.hdr.include("portenvironment.hpp");
+
```
joakim-brannstrom: Should we add the source code for the runtime environment?
It would make it possible to create integration tests.
Extremely useful for regression testing.

joakim-brannstrom: The file structure is partially, intentionally structured to allow this.
<plugin_name>/source/ <- D code
add freely other directories with different content.

ploq: As of right now we are only rewriting all of our old code and trying to make it more maintainable and easier to read. Would be nice though

plugin/fuzz/source/dextool/plugin/fuzz/sutxml.d
```diff
@@ -0,0 +1,208 @@
...
+                ~ "::" ~ baseName(filename, ".xml");
+            //dfmt on
+    }
+
+    @trusted Array!XMLType getAllSubtypes(Element types) { //Types = Types tag
+        Array!XMLType xml_types;
+        foreach (Element subtype; types.elements) {
+            XMLType curr_type = XMLType(subtype.tag.attr["name"], subtype.tag.attr["type"]);
+
+            if (subtype.tag.attr.get("min", "-1NOTFOUND") != "-!NOTFOUND") {
```
joakim-brannstrom: If I see correctly attr is an associative array.
The following pattern ensures only one lookup is made in the AA:
```d
string[string] arr = ["a": "foo", "b":"borf"];
if (auto v = "a" in arr) {
    writeln(*v);
}
```

plugin/fuzz/source/dextool/plugin/fuzz/xml_parse.d
```diff
@@ -0,0 +1,255 @@
...
+
+    string name;
+    string type;
+    string unit; ///Optional
+    string min; ///Optional
+    string max; ///Optional
+}
+
+struct ContinuesInterface {
+    int direction; //From = 0; To = 1;
```
joakim-brannstrom: Use an enum. This is very magic ;)

plugin/fuzz/source/dextool/plugin/fuzz/sutxml.d
```diff
@@ -186,7 +186,15 @@ public:
     }
 }
 
+void _debug() {
+    writeln(SutEnvironment("./sut_unittest/namespaces/").getAllNamespaces);
```
joakim-brannstrom: Tip, there exist a debug functionality i the standard library. It allows runtime configuration of debug level.
```d
import logger = std.experimental.logger;
logger.info(SutEnvironment("./sut_unittest/namespaces/").getAllNamespaces);
```

plugin/fuzz/source/dextool/plugin/fuzz/sutxml.d
```diff
@@ -186,7 +186,15 @@ public:
...
+    writeln(SutEnvironment("./sut_unittest/namespaces/").getAllXMLInterfaces);
+    SutEnvironment("./sut_unittest/namespaces/").getTypesFromInterfaces;
+}
+
 int main() {
+    debug import std.stdio;
+    debug _debug();
```
joakim-brannstrom: nice, using "debug" :)

source/cpptooling/data/representation.d
```diff
@@ -1423,6 +1423,20 @@ const:
...
+    this(CppNs name, CppNsStack stack, CppClass[] classes, CFunction[] funcs, CppNamespace[] namespaces, CxGlobalVariable[] globals) {
+	    this.name_ = name;
+        this.stack = stack;
+        /*this.classes = classes;
+        this.funcs = funcs;
+        this.globals = globals;*/
+	    //this.stack = CppNsStack(stack.dup);
+    } 
+
+    CppNamespace dup() {
```
ploq: This is probably a bug in the standard CppTestdouble plugin.

resideInNs() and equivalent functions does not work as expected, only returns top namespace (bar in foo::bar)

ploq: This should not be considered as a fix!

joakim-brannstrom: /thinking

joakim-brannstrom:
Nice, travis going green :)

joakim-brannstrom:
I realize I don't have your mail addresses here at home. My intentions are to drop by tomorrow afternoon. Lets hope I don't get lost in the labyrinth ;)

plugin/fuzz/source/dextool/plugin/fuzz/backend/fuzzvariant.d
```diff
@@ -407,40 +398,43 @@ body {
...
-                            ns.resideInNs[$ - 1].payload, nss);
-                    }
-                    b.visit!((const CppMethod a) => generateCppMeth(a,
-                        classes[fqn_class], class_name, fqn_class, xmlp),
-                        (const CppMethodOp a) => writeln(""),
-                        (const CppCtor a) => generateCtor(a, inner),
-                        (const CppDtor a) => generateDtor(a, inner));
-                }
+        foreach (a; ns.classRange) {
+            string class_name = a.name[2 .. $];
```
joakim-brannstrom: 2 ... $, it is to remove "I_"?

ploq: The commit you were reviewing was old, so that part of the code has been removed from the latest version, but yes that was ti remove I_

plugin/fuzz/source/dextool/plugin/fuzz/backend/fuzzvariant.d
```diff
@@ -460,7 +454,7 @@ body {
     import std.string : toLower, indexOf;
     import std.algorithm : endsWith;
 
-    string base_class;
+    string base_class = "";
```
ploq: As said in previous comment, generateClass() has been rewamped with use of tagging and splitted to three functions instead of one large.

source/cpptooling/data/representation.d
```diff
@@ -1381,14 +1382,6 @@ const:
...
 
-/// Dictates how the namespaces are merged.
-enum MergeMode {
-    /// Merge everything except nested namespaces.
-    shallow,
-    /// Merge everything.
-    full
-}
-
 @safe struct CppNamespace {
```
ploq: Something broke with the newer representation.d, we should look into what the new one does differently. Only generates one namespace with the new version. 
(Probably something with dup)

joakim-brannstrom: What are the tests that you ran that started to fail after the merge?
I'm thinking of looking into the problem but need to be able to reproduce it.

ploq: Did not run any tests unfortunately, buuuuuuut I found the problem. I believe that we are using put(CppNamespace) in CppRoot wrongly. 

In our process() function we are creating a new CppRoot to put all our namespaces that we filtered out (line 98 in backend/fuzzvariant.d), either something broke in the put function or we should be doing something else.

ploq: Seems like it automatically merges every namespace that already exists (first namespace) and ignores the rest?

ploq: Naaa, found the problem, every namespace that rawFilter returns has the same id

joakim-brannstrom:
Great you found it :)

plugin/fuzz/source/dextool/plugin/fuzz/backend/fuzzvariant.d
```diff
@@ -626,45 +627,52 @@ import cpptooling.data.type;
 			    string min = minmax["min"];
 			    string max = minmax["max"];
 			    string type_type = minmax["type"];
-                string type_ns = minmax["namespace"];
+			    string type_ns = minmax["namespace"];
 			    
 			    final switch (type_type) {
```
joakim-brannstrom: final has to meaning here. might as well remove it.
It is useful in those cases one switch over an enum time to guarantee at compile time that all enum literals are handled.

plugin/fuzz/source/dextool/plugin/fuzz/backend/fuzzvariant.d
```diff
@@ -592,27 +592,28 @@ import cpptooling.data.type;
...
+    import std.format : format;
 
     with (inner_class) {
         with (private_) {
             logger.trace("class_name: " ~ class_name);
             logger.trace("generateClass fqn_ns: " ~ fqn_ns);
             foreach (ciface; ns.interfaces.ci) {
-                stmt(E(fqn_ns ~ "::" ~ ciface.name ~ "T " ~ ciface.name.toLower));
+		string expr = format("%s::%sT %s", fqn_ns, ciface.name, ciface.name.toLower);
+                stmt(E(expr));
```
joakim-brannstrom: Maybe simplify even further by doing e.g.:
```d
stmt(format(.....));
```

plugin/fuzz/source/dextool/plugin/fuzz/backend/fuzzvariant.d
```diff
@@ -626,45 +627,52 @@ import cpptooling.data.type;
...
-                string type_ns = minmax["namespace"];
+			    string type_ns = minmax["namespace"];
 			    
 			    final switch (type_type) {
 			    case "SubType":
-				stmt(E(ciface.name.toLower ~ "." ~ ditem.name) = E(
-									   `randomGenerator->generate("` ~ type ~ ` ` ~ ciface.name
-									   ~ ` ` ~ ditem.name ~ `", `~min~`, `~max~`)`));
+				string var = format("%s.%s", ciface.name.toLower, ditem.name);
+				string expr = format("randomGenerator->generate(\"%s %s %s\", %s, %s)",
```
joakim-brannstrom: The backtick, `, can also be used for strings. Because it is uncommon in "other" languages it means it almost always "just works".

plugin/fuzz/source/dextool/plugin/fuzz/xml_parse.d
```diff
@@ -0,0 +1,457 @@
...
+import std.array;
+import std.container.array;
+import std.file;
+import std.path;
+import std.string;
+import std.xml;
+import std.typecons;
+
+import std.stdio;
+
```
joakim-brannstrom: I recommend making the whole module "@safe".
It will _probably_ allow most of the other code to be safe too.

To make the destructor safe in the structs replace all usage of Array with the builtin arrays.
```d
Arra!Primitive prims;
// by
Primitive[] prims;
```

The insertBack is then replace by a ~=.

Why? Then the following bug would have been found by the compiler (see [1]).

plugin/fuzz/source/dextool/plugin/fuzz/xml_parse.d
```diff
@@ -0,0 +1,457 @@
...
+    }
+
+    Namespace[string] parseBaseDir() {
+        import std.stdio;
+
+        Namespace[string] namespaces;
+        Global glob;
+        foreach (string xml_file; xml_files) {
+            string curr_ns = namespace(xml_file);
+            auto doc = new Document(cast(string) std.file.read(xml_file));
```
joakim-brannstrom: [1] This cast is very unsafe. It is forcing a void[] to a string[] which will lead to e.g. UTF-8 decoding problems.
What needs to be done is a read -> validation -> cast.

joakim-brannstrom: This may not be the best way, but it is a way. Please be critical and if you find a better way, tell me.
```d
import std.utf : validate;
string raw = () @trusted { return cast(string) std.file.read(xml_file); }();
validate(raw);
```

plugin/fuzz/source/dextool/plugin/fuzz/xml_parse.d
joakim-brannstrom: whole module, by adding the following after the import.
```d
@safe:
```

plugin/fuzz/source/dextool/plugin/fuzz/xml_parse.d
ploq: https://dlang.org/phobos/std_file.html#.readText readText() seems to suite our needs quite well, basically the same thing

joakim-brannstrom: Yes seems like that. Internally using validate :)
Thank you.

ploq: Too bad that the XML docs doesn't show this

https://dlang.org/phobos/std_xml.html (the first examples)

plugin/fuzz/source/dextool/plugin/fuzz/frontend/fuzz.d
```diff
@@ -0,0 +1,250 @@
...
+    string res;
+    writeln(hfiles);
+    //
+    foreach(hfile, cmd; hfiles) {
+        auto db_search_result = variant.getCompileDB.appendOrError(user_cflags, cmd[1],
+                CompileCommandFilter(defaultCompilerFlagFilter, 1));
+
+        use_cflags = db_search_result.get.cflags;
+
+        auto ctx = ClangContext(Yes.useInternalHeaders, Yes.prependParamSyntaxOnly);
```
joakim-brannstrom: tip: the ClangContext can be moved outside of the loop. Make it a little bit more efficient.

plugin/fuzz/source/dextool/plugin/fuzz/frontend/fuzz.d
```diff
@@ -0,0 +1,250 @@
...
+    //auto range = variant.getCompileDB.map!(a => a.absoluteFile).enumerate;
+
+    string res;
+    writeln(hfiles);
+    //
+    foreach(hfile, cmd; hfiles) {
+        auto db_search_result = variant.getCompileDB.appendOrError(user_cflags, cmd[1],
+                CompileCommandFilter(defaultCompilerFlagFilter, 1));
+
+        use_cflags = db_search_result.get.cflags;
```
joakim-brannstrom: tip: always chech ".isNull" before calling get on a Nullable. Or nasty things will happen.

joakim-brannstrom: For those cases where it is intended to not check it a "reassuring comment" is useful :)

plugin/fuzz/source/dextool/plugin/fuzz/frontend/fuzz.d
```diff
@@ -13,6 +13,35 @@ import dextool.compilation_db;
...
 import xml_parse;
 
+//Search and get headers from specific file
+auto getHeaderFiles(CompileCommandDB compile_db, CompileCommand file) @safe {
+    import std.range : array;
+    import std.algorithm : uniq, filter, map, sort;
+    import std.file : dirEntries, SpanMode;
+    import std.conv : to;
+
+    string[] flags = parseFlag(file, CompileCommandFilter()).filter!(a => a[0] != '-').array;
```
joakim-brannstrom: tip: change CompilerCommandFilter to defaultCompilerFilter.
why? clang and gcc do not support the same flags. The default implementation of filter, defaultCompilerFilter, covers the most common deviations. It is empirically derived through trial and error.

plugin/fuzz/source/dextool/plugin/fuzz/backend/fuzzvariant.d
```diff
@@ -270,7 +270,7 @@ final class FuzzVisitor(RootT, ProductT) : Visitor {
             v.accept(visitor);
 
             root.put(visitor.root);
-            container.put(visitor.root, visitor.root.fullyQualifiedName);
+            //container.put(visitor.root, visitor.root.fullyQualifiedName);
```
joakim-brannstrom: Yes, old technical debt that got fixed. It is up to the plugins to store the CppClass'es if they are needed

ploq:
@joakim-brannstrom What do you think about this? https://github.com/test-pipeline/orthrus

joakim-brannstrom:
Looks nice. Try it. The one thing that caught my eye is the dependency on autotools.

ploq:
True, if there were enough time, a fork of that repo and changing compilers and flags from variables would make it work with the current output

PR #344, opened 2017-04-24
Cmake mac fixes

ploq:
A new build equals some specific Mac bugs...
findLibClang.cmake
* Added new path for libclang
* Added Apple check for unix. if(UNIX) is not enough since Macs are identified as UNIX also.
* Removed universal --no-as-needed since Mac does not support this option. (sigh)
* Splitted rpath for Macs into two arguments

common.cmake
* ln does not have the T flag on mac, removed for macs (sigh)

source/application/CMakeLists.txt
* cp does not have --reflink on mac, removed for macs (sigh)


cmake/Modules/FindLibClang.cmake
```diff
@@ -53,14 +54,14 @@ if(NOT LIBCLANG_LDFLAGS)
     # -rpath is relative path for all linked libraries.
     # The second "." is argument to rpath.
 
-    if(UNIX)
-        set(LIBCLANG_LDFLAGS_OS "-L--enable-new-dtags -L-rpath=.")
-    elseif(APPLE)
-        set(LIBCLANG_LDFLAGS_OS "-L-rpath=${LIBCLANG_DIR_PATH}")
+    if(UNIX AND NOT APPLE)
```
joakim-brannstrom: UNIX is always true on MacOSX?
From what I can see it should have equivalent behavior if the checks are flipped, first apple then unix. The "AND..." can be removed

cmake/Modules/FindLibClang.cmake
```diff
@@ -53,14 +54,14 @@ if(NOT LIBCLANG_LDFLAGS)
...
+        set(LIBCLANG_LDFLAGS_OS "-L--enable-new-dtags -L-rpath=. -L--no-as-needed -L-l:${LIBCLANG_LIB}")
+    elseif(UNIX AND APPLE)
+        set(LIBCLANG_LDFLAGS_OS "-L-rpath -L${LIBCLANG_DIR_PATH} -L-lclang")
```
joakim-brannstrom: rpath missing arguments? meant to be -L-rpath=${LIBCLANG_DIR_PATH}?

joakim-brannstrom: The -L${LIBCLANG_DIR_PATH} is redundant, already covered by the flag set at line 64

ploq: No it is meant to be two arguments. ld on Mac does not support rpath=, and therefore I split them into two arguments which is then parsed correctly by ld.

cmake/common.cmake
```diff
@@ -34,12 +34,21 @@ endfunction()
...
         TARGET ${name}
         POST_BUILD
         COMMAND ln -sfT ${CMAKE_CURRENT_BINARY_DIR}/${name} ${CMAKE_BINARY_DIR}/${name}
         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
     )
+   elseif(UNIX AND APPLE) 
+    add_custom_command(
+        TARGET ${name}
+        POST_BUILD
+        COMMAND ln -sf ${CMAKE_CURRENT_BINARY_DIR}/${name} ${CMAKE_BINARY_DIR}/${name}
```
joakim-brannstrom: hmm I see. I think it may be simplified even further.
Instead of adding the branch what do you think about:
 - add a before the ln to remove the symlink, if it exist. COMMAND rm -f ${CMAKE_BINARY_DIR}/${name}
 - remove the "T" from ln.

Sorry for the T, didn't know it was GNU ln specific.

cmake/Modules/FindLibClang.cmake
ploq: Yup, UNIX is always true on Mac, to differentiate them one would need to check APPLE too

source/application/CMakeLists.txt
```diff
@@ -17,12 +17,21 @@ build_d_executable(
...
     TARGET ${DEXTOOL_MAIN_EXE}
     POST_BUILD
     COMMAND cp --reflink=auto ${CMAKE_CURRENT_BINARY_DIR}/${DEXTOOL_MAIN_EXE} ${CMAKE_BINARY_DIR}/${DEXTOOL_MAIN_EXE}
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
 )
+elseif(UNIX AND APPLE)
+add_custom_command(
+    TARGET ${DEXTOOL_MAIN_EXE}
+    POST_BUILD
+    COMMAND cp ${CMAKE_CURRENT_BINARY_DIR}/${DEXTOOL_MAIN_EXE} ${CMAKE_BINARY_DIR}/${DEXTOOL_MAIN_EXE}
```
joakim-brannstrom: remove the branch, remove the --reflink=auto. I just tried to be "smart". lets keep it simple :)

joakim-brannstrom:
sorry for the differences. thank you for the pull request and the time to make it work.

cmake/Modules/FindLibClang.cmake
joakim-brannstrom: Ohh you mean. dmd is called with: -L-rpath -Lfoo
dmd passes it on to the linker as: ld -rpath foo
?

ploq: Yes exactly! :)

cmake/common.cmake
ploq: Hmm, yes that seems resonable! Will add this tomorrow

source/application/CMakeLists.txt
ploq: Will fix tomorrow :)

cmake/Modules/FindLibClang.cmake
joakim-brannstrom: cool, then I understand. keep as is :)

Issue #347, opened 2017-04-25
No enums in dsrcgen?


ploq:
Am I missing something or doesn't enums exist in dsrcgen?

It should :)

joakim-brannstrom:
I agree :)

What do you think about #348 ?

ploq:
Yes, seems good!

PR #348, opened 2017-04-25
dsrcgen: add support for C enum and C++ enum class

joakim-brannstrom:


joakim-brannstrom:
Merged in response to #347

PR #362, opened 2017-04-29
doc: add plugin example

joakim-brannstrom:

PR #363, opened 2017-04-30
doc: fix plugin example

joakim-brannstrom:


joakim-brannstrom:
Updated after feedback on #362

joakim-brannstrom:
@ploq You had some comments earlier in the project that deXtool was lacking documentation. I'm trying bit by bit to fix it.
I'll gladly take any feedback that you can provide :)

The example do not show how to use "representation.d" but... one thing at a time ;)

ploq:
Hmm... Well I don't think that representation.d needs much documentation in the example plugin (not showing every statement etc.). It is quite self explanatory how it works if one would implement a for-loop, some statements and maybe some C++ stuff like templates. Basically some basic functionality about dsrcgen. 

Other than that, I think that the example plugin is good enough. :)

joakim-brannstrom:
Thanks for the feedback :)
Updated with a bit more examples of using the analyze and dsrcgen.

PR #373, opened 2017-05-06
test: missing that dextool binaries must be built before

joakim-brannstrom:
integration tests can be executed.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -18,6 +18,9 @@ This gives access to the make target _test_.
 
 To run the tests:
 ```sh
-make check # runs unittests
+# builds and runs the unittests
+make check
+
+# builds and runs the integration tests
 make check_integration
 ```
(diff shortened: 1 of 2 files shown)


joakim-brannstrom:
@christoffer-nylen  better?

DEVELOPER.md
```diff
@@ -18,6 +18,12 @@ This gives access to the make target _test_.
...
 To run the tests:
 ```sh
-make check # runs unittests
+# builds and runs the unittests
+make check
+
+# Integration tests require that the dextool binaries are built
+make all
+
+# builds and runs the integration tests
```
christoffer-nylen: If ``make all`` builds everything, what gets built in this step?

joakim-brannstrom:
Adjusted the build system to make "check integration" depend on "make all". No extra documentation needed then. It will "just work" as expected.

Issue #392, opened 2017-06-13
ctestdouble/cpptestdouble does not correctly expand tilde


ploq:
When giving the in-flag a tilde it does not get expanded to /home/username (the tilde is never removed).
This ONLY occurs when having an equal sign after in. I guess that this has to do with bash not expanding the tilde (?). 

For example:
Input: `--in=~/something/something.hpp`
Output: `/home/username/~/something/something.hpp`

ploq:
https://dlang.org/library/std/path/expand_tilde.html should probably be used with in_file as the argument in genCpp() and its equal in ctestdouble

PR #393, opened 2017-06-14
c_cpptestdouble: expand tilde in paths

joakim-brannstrom:


joakim-brannstrom:
Should solve the issue in #392  :)

PR #400, opened 2017-06-21
cpptooling: improve search algorithm of comp db

joakim-brannstrom:


joakim-brannstrom:
@ploq this may be of interest for you

PR #447, opened 2017-09-10
llvm_hiwrap: high level wrapper for the llvm-c API

joakim-brannstrom:


joakim-brannstrom:
this is just to start using the CI server

PR #601, opened 2018-02-26
mutate: (doc) update README

joakim-brannstrom:


plugin/mutate/README.md
```diff
@@ -1,78 +1,105 @@
...
-Reconfigure and prebuild with the tests activated:
+Suppose the source code is in ... and the test code is elsewhere. Use the ``--restrict`` option to specify what to analyze:
 ```sh
-cmake -Dgtest_build_tests=ON -Dgmock_build_tests=ON ..
-make
+dextool mutate analyze --compile-db compile_commands.json --restrict .. -- -D_POSIX_PATH_MAX=1024
 ```
 
-Create a file tester.sh with this content:
+Create a script ``tester.sh`` that runs the entire test suite when invoked:
```
joakim-brannstrom: maybe add what I mentioned to you. That it is important that compiler.sh build all code, both test and program/SUT. it is important that test.sh only execute the tests.
Otherwise two things happens:
- many mutants will be wrongly classified as timeout when it was just the compilation that change a lot. GoogleTest is interesting in this case. Change a template and the compilation time may change to 1-2 minutes + 20s execution time of tests. Change a *hidden* implementation (.cpp) that only affects one translation unit and the compilation time is around a couple of seconds + 20s execution time.
- the mutation tester is unable to classify a mutation as "killed by compiler". it would be classified as "killed by the test suite". this results in a higher mutation score than it actually is. This is because the mutation score (for now) is calculated as survived / (killed+timeout+survived). "killed by compiler" is NOT part of the total because those are just invalid constructs. The programmer is never able to even write such code which goes against the assumption of a competent programmer that try to implement according to a specification.

plugin/mutate/README.md
```diff
@@ -1,78 +1,105 @@
...
-It is possible to run multiple `test` against the same database.
-Just make sure they don't mutate the same source code.
-
-To see the result:
+Generate the mutation testing result:
 ```sh
 dextool mutate report --restrict .. --level alive --mutant lcr
 ```
 
-## Compiling Google Test with Coverage
```
joakim-brannstrom: Maybe keep the headline to "separate" this chapter from the "mutation testing"?
It is technically _not_ part of the mutation testing but something that is probably interesting for the reader to look at.

plugin/mutate/README.md
```diff
@@ -1,78 +1,105 @@
...
 ```
 
-To generate a HTML coverage report:
+Generate a HTML coverage report with:
 ```sh
 lcov -c --gcov-tool /usr/bin/gcov -d . --output-file app.info
 genhtml app.info -o html
 ```
+
+# Administration
```
joakim-brannstrom: this chapter should probably be rewritten to highlight the use of `dextool mutate admin`.
The most interesting aspect would be how to reset _survived_ mutants to retest those when a test case has been added to the SUT.

plugin/mutate/README.md
```diff
@@ -1,78 +1,105 @@
...
-Reset all mutations of a kind to unknown which forces them to be tested again:
-```sh
-sqlite3 dextool_mutate.sqlite3 "update mutation SET status=0 WHERE mutation.kind=FOO"
-```
+Dextool's plugin for mutation testing of C/C++ projects. It can help you design new tests and evaluate the quality of existing  tests.
+
+### Features
+
+TODO: List supported mutation operators etc.
+TODO: What are the selling points etc?
```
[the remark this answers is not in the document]
christoffer-nylen: Great summary!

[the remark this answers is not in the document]
joakim-brannstrom: Now when I think about it. The timeout handling could be said that it handles a "flaky" test suite hmm.

plugin/mutate/README.md
```diff
@@ -1,78 +1,116 @@
...
+Dextool's plugin for mutation testing of C/C++ projects. It can help you design new tests and evaluate the quality of existing  tests.
+
+### Features
+
+* Provides support for conventional mutation operators: AOR, ROR, DCC, DCR, LCR, COR.
+* Can continue from where a testing session was interrupted.
+* Allows multiple instances to be run in parallel.
+* Makes type aware ROR to reduce the number of equivalent mutants.
+* Can reuse previous results when a subset of the SUT changes by only testing those changes (files for now).
+* Provides multiple report formats.
```
joakim-brannstrom: maybe at at the end (Markdown, compiler warnings, JSON)

joakim-brannstrom:
Any progress?

PR #667, opened 2018-06-26
Hello world-plugin

nils-petersson:
Simple plugin for finding for-loops in files


plugin/eqdetect/source/dextool/plugin/eqdetect/runner.d
```diff
@@ -0,0 +1,105 @@
...
+
+    import std.file;
+    import std.conv : to;
+    import std.algorithm : canFind;
+
+    File file = File(to!string(pargs.file), "r");
+    int i = 1;
+    while (!file.eof()) {
+        string s = file.readln();
+        if(canFind(s, "for") && !canFind(s, "//")){
```
joakim-brannstrom: creative solution. But I would suggest to use the AST to familiarize with the visitor and clang.

plugin/eqdetect/source/dextool/plugin/eqdetect/runner.d
```diff
@@ -0,0 +1,105 @@
...
+        return ExitStatusType.Errors;
+    }
+
+    import std.file;
+    import std.conv : to;
+    import std.algorithm : canFind;
+
+    File file = File(to!string(pargs.file), "r");
+    int i = 1;
+    while (!file.eof()) {
```
joakim-brannstrom: can replace:
```d
File file = ...
while (!file.eof()) {
   string s = file.readln();
```
with:
```d
foreach (s; File(pargs.file).byLine) {....
```

joakim-brannstrom: which could be compressed to a one liner if using count from std.algorithm :)

joakim-brannstrom:
The CI fail because the counter of installed files need to be increased. Change 15 to 16 in `tools/travis_test_install.sh`

joakim-brannstrom:
Nice improvement. Good job :)

joakim-brannstrom:
Good progress. Nice to see the expanded knowledge if both D and dextool.

joakim-brannstrom:
interesting. we should really generalize the DB integration so it is reusable between your plugin and mutate.
The files schema.d, standalone.d and type.d is intended to be this "standalone" part. I just haven't gotten around to move fully refactor them into a independent library.

plugin/eqdetect/source/dextool/plugin/eqdetect/runner.d
```diff
@@ -57,9 +57,12 @@ ExitStatusType runPlugin(string[] args) {
 
         exit_status = analyzeFile(AbsolutePath(FileName(m.path)), cflags, visitor, ctx);
         import std.path : baseName;
-        import dextool.plugin.eqdetect.subfolder : writeToFile;
-        writeToFile(visitor.generatedSource.render, baseName(m.path), m.kind, m.id, false);
-        writeToFile(visitor.generatedMutation.render, baseName(m.path), m.kind, m.id, true);
+        import dextool.plugin.eqdetect.subfolder : writeToFile, SnippetFinder;
+        string source_path = writeToFile(visitor.generatedSource.render, baseName(m.path), m.kind, m.id, "_source_");
```
joakim-brannstrom: if it is a path I would recommend to use the struct AbsolutePath or variants of it. It ensures that the path is **always** absolute. It makes it easier to pass them around in the program.

`source/dextool/type.d`

joakim-brannstrom: it also avoid extra work such as at other places in the program checking if it is a path. You can see it from the opAssign overloads and constructors.

In this case it could be written as:
`auto source_p = ....Path.AbsolutePath`

plugin/eqdetect/source/dextool/plugin/eqdetect/runner.d
```diff
@@ -57,9 +57,12 @@ ExitStatusType runPlugin(string[] args) {
...
+        string mutant_path = writeToFile(visitor.generatedMutation.render, baseName(m.path), m.kind, m.id, "_mutant_");
+        auto s = SnippetFinder.generateKlee(visitor.function_params, source_path,
+        mutant_path, visitor.function_name);
+        writeToFile(s, baseName(m.path), m.kind, m.id, "_klee_");
         writeln("---------------------------------------");
```
joakim-brannstrom: tip: use the logging framework. It makes it easier to control how much info is printed.

plugin/eqdetect/source/dextool/plugin/eqdetect/subfolder/codegenerator.d
```diff
@@ -51,4 +51,48 @@ class SnippetFinder{
         temp = temp ~ content[mutation.offset_end .. content.length];
         return temp;
     }
+
+    @trusted static auto generateKlee(string[] params, string source_name, string mutant_name, string function_name){
+        import std.stdio;
+        import std.format;
+        //string code;
+        import dsrcgen.c;
```
joakim-brannstrom: tip: there is a cpp module too :)

plugin/eqdetect/source/dextool/plugin/eqdetect/subfolder/codegenerator.d
```diff
@@ -51,4 +51,48 @@ class SnippetFinder{
...
+        auto code = new CModule();
+        //add klee imports
+        code.include("<klee/klee.h>");
+        code.include("<assert.h>");
+
+        //add import for files under test
+        code.include(format("\"%s\"", source_name));
```
joakim-brannstrom: tip: in D it is possible to use the backtick too. It usually avoids these escaping.

```d
format(`foo "%s"`, source_name);
```

plugin/eqdetect/source/dextool/plugin/eqdetect/subfolder/codegenerator.d
```diff
@@ -51,4 +51,48 @@ class SnippetFinder{
...
+
+        string func_params;
+        //add klee-main
+        with(code.func_body("int", "main")){
+            //variable declaration
+            for(int i = 0; i < params.length; i++){
+                stmt(format("%s var%s;", params[i], i));
+                func_params = func_params ~ format("var%s,", i);
+            }
+            func_params = func_params[0..$-1];
```
joakim-brannstrom: Removing the last one because? (maybe a comment if it is intentional)

plugin/eqdetect/source/dextool/plugin/eqdetect/subfolder/codegenerator.d
```diff
@@ -51,4 +51,48 @@ class SnippetFinder{
...
+
+        //add import for files under test
+        code.include(format("\"%s\"", source_name));
+        code.include(format("\"%s\"", mutant_name));
+
+        string func_params;
+        //add klee-main
+        with(code.func_body("int", "main")){
+            //variable declaration
+            for(int i = 0; i < params.length; i++){
```
joakim-brannstrom: tip: nifty loops as in python works here.
```d
foreach (const i; 0 .. params.length) {...}
```

plugin/eqdetect/source/dextool/plugin/eqdetect/subfolder/codegenerator.d
peatersson: Yes, we want to use it in the future. There are currently a lot of code written only to get something to work with KLEE that needs changing or completely removal in the future.

plugin/eqdetect/source/dextool/plugin/eqdetect/subfolder/codegenerator.d
joakim-brannstrom: if it is iterating over the elements and you need an index too then this works:
```d
foeach (i, v; params) {
   stmt(format("%s var%s", v, i);
}
```

plugin/eqdetect/source/dextool/plugin/eqdetect/subfolder/codegenerator.d
```diff
@@ -51,4 +51,48 @@ class SnippetFinder{
...
+        code.include("<klee/klee.h>");
+        code.include("<assert.h>");
+
+        //add import for files under test
+        code.include(format("\"%s\"", source_name));
+        code.include(format("\"%s\"", mutant_name));
+
+        string func_params;
+        //add klee-main
+        with(code.func_body("int", "main")){
```
joakim-brannstrom: missing argv and argc?

plugin/eqdetect/source/dextool/plugin/eqdetect/subfolder/codegenerator.d
```diff
@@ -51,4 +51,48 @@ class SnippetFinder{
         temp = temp ~ content[mutation.offset_end .. content.length];
         return temp;
     }
+
+    @trusted static auto generateKlee(string[] params, string source_name, string mutant_name, string function_name){
```
joakim-brannstrom: After reading the function I do not fully understand its intent.
I think 2-3 lines of comments explaining the intent such as what the code in main intends to do.

joakim-brannstrom: tip: create types for the parameters to add that the programming language can understand beyond "string" and the name.
mutant_name is strictly different from function_name.

source_name could be an AbsolutePath.
mutant_name could be a:
```d
struct MutantName {
   string value;
   alias value this;
}
```
function_name could be a CFunctionName from source/cpptooling/type.d.

This gives you clarity, ability to see in the type system *how* the data is used, easier to refactor etc.

tip: Kevlin Henneys video about stringly typed systems. [ Kevlin Henney - Seven Ineffective Coding Habits of Many Programmers 26min+](https://www.youtube.com/watch?v=ZsHMHukIlJY)

plugin/eqdetect/source/dextool/plugin/eqdetect/subfolder/codegenerator.d
```diff
@@ -51,4 +51,48 @@ class SnippetFinder{
...
+            }
+            func_params = func_params[0..$-1];
+            //symbolic variable for each parameter
+            for(int i = 0; i < params.length; i++){
+                stmt(format("klee_make_symbolic(&var%s, sizeof(%s), \"var%s\")", i, params[i], i));
+            }
+            //equivalence detection
+            import std .path;
+            with(if_(format("%s.%s == %s.%s",
+                stripExtension(source_name), format("%s(%s)",function_name, func_params),
```
joakim-brannstrom: double format? :)
maybe merge into one format?

plugin/eqdetect/source/dextool/plugin/eqdetect/subfolder/codegenerator.d
```diff
@@ -51,4 +51,48 @@ class SnippetFinder{
...
+                stripExtension(mutant_name), format("%s(%s)",function_name, func_params)))){
+
+                return_("1");
+            }
+            with(else_){
+                stmt("klee_assert(0);");
```
joakim-brannstrom: tip: can remove the trailing ";". stmt ensure they always have a terminator token.

plugin/eqdetect/source/dextool/plugin/eqdetect/subfolder/dbhandler.d
```diff
@@ -83,8 +83,9 @@ class DbHandler{
         auto stmt = db.prepare(format("SELECT path, lang FROM files WHERE id='%s';", file_id));
         auto res = stmt.execute;
 
+        import std.path;
         auto path = res.front.peek!string(0);
-        path = "../" ~ path; //Ugly solution for now
+        path = buildPath("../" ~ path); //Ugly solution for now
```
joakim-brannstrom: nice. Further refinement.
tip:
```d
buildPath("..", path);
```

It concatenates paths "correctly". Especielly useful to avoid extra "/" and make it platform independent.

plugin/eqdetect/source/dextool/plugin/eqdetect/subfolder/filewriter.d
```diff
@@ -23,14 +23,11 @@ static void writeToFile(string text_to_write, string base, int kind, int id, boo
...
     string filename;
 
-    if (!isMutation){
-        filename = dir ~ "/" ~ stripExtension(base) ~ "_source_" ~ to!string(id) ~ "_"
-        ~ to!string(cast(mutationStruct.Kind)kind) ~ extension(base);
-    } else {
-        filename = dir ~ "/" ~ stripExtension(base) ~ "_mutation_" ~ to!string(id) ~ "_"
-        ~ to!string(cast(mutationStruct.Kind)kind) ~ extension(base);
-    }
+    filename = dir ~ "/" ~ stripExtension(base) ~ filetype ~ to!string(id) ~ "_"
```
joakim-brannstrom: tip: buildPath

plugin/eqdetect/source/dextool/plugin/eqdetect/runner.d
peatersson: Yes, the plugin are writing a lot of information that needs to be changed. Thank you for tip!

plugin/eqdetect/source/dextool/plugin/eqdetect/runner.d
```diff
@@ -63,9 +63,23 @@ ExitStatusType runPlugin(string[] args) {
...
+        import std.file: getcwd;
+        import std.format : format;
+
+        auto cwd = getcwd();
+
+        writeln(executeShell(format("docker run -it --name=klee_container -v %s:/home/klee/mounted klee/klee mounted/klee.sh", cwd)).output);
+        executeShell("docker rm klee_container");
+        executeShell("rm -rf eqdetect_generated_files/*");
+
         writeln("---------------------------------------");
```
peatersson: This solution is only temporary since running KLEE via docker is not the end goal. The problem with starting a script via a docker that compiles and runs KLEE is that it needs sudo privileges (as it is implemented now). This causes several issues, but will work for the moment when we are testing and continuously changing the implementation.

joakim-brannstrom: Tip: if you arent interested in the captured output spawnShell may be used.
spawnShell(....).wait;

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -155,6 +155,16 @@ immutable mutation_tbl = "CREATE TABLE %s (
     FOREIGN KEY(mp_id) REFERENCES mutation_point(id) ON DELETE CASCADE
     )";
 
+immutable mutation4_tbl = "CREATE TABLE %s (
+    id      INTEGER PRIMARY KEY,
+    mp_id   INTEGER NOT NULL,
+    kind    INTEGER NOT NULL,
+    status  INTEGER NOT NULL,
+    time    INTEGER,
+    eq      INTEGER,
```
joakim-brannstrom: Do we want to ensure eq is never null? What is the default state? unknown == 0?

peatersson: No, we want it to be set to something all the time. We are thinking 
- non-eq (klee found a testcase to kill the mutant)
- eq (all paths traversed and no one violated the assertion)
- running (could be redundant)
- unknown (not tested, this could be the standard)
- timed out (klee timed out)

joakim-brannstrom: Looking good. Yes running is probably not needed for now. It can be added in the future as an optimization for work distribution, if needed.

joakim-brannstrom: For the non-eq it would be really really reaaaally useful to give test data to the user. Help the user kill a mutant. Or help to understand what logic is redundant

joakim-brannstrom: A side question to think about. How can the timeout be automatically derived?

peatersson: Yes, we are currently doing the parsing from KLEE, where we can obtain test data, so we will include that. 

We were thinking of using the option flags for KLEE, where we can set how long KLEE get before halting execution. This could be dynamically done, i.e. user input or similar, but will probably be set to something like 30 seconds for now. If KLEE halts execution, we can know this by looking for a specific file that KLEE outputs.

joakim-brannstrom:
Nice cleanup

PR #794, opened 2018-12-16
doc: Add build.sh to mutate readme

christoffer-nylen:


plugin/mutate/README.md
```diff
@@ -69,6 +69,14 @@ Generate a database of all mutation points:
 dextool mutate analyze
 ```
 
+Create a file `build.sh` that will build the subject under test when invoked:
```
joakim-brannstrom: Instead of adding a new entry change name of compile.sh further down in the readme.

PR #797, opened 2018-12-16
mutate: inform the user what database is being used

joakim-brannstrom:


joakim-brannstrom:
I hope this helps the user to understand why the report may be empty because the wrong database where used.

PR #799, opened 2018-12-17
mutate: (html) change color for killed by compiler

christoffer-nylen:
Test cases can not affect mutation points that are killed by the compiler. The idea of the selected color ([#eeeeee](https://www.color-hex.com/popular-colors.php)) is to indicate that the mutation result for such mutants can be disregarded in the context of test effectiveness.


christoffer-nylen:
here's an where the gray color is used [example](https://imgur.com/a/HdjSqYm)

PR #831, opened 2019-03-21
mutate: mutation status results are wiped when copying DB from jenkins

joakim-brannstrom:
This is hard to replicate but I think the problem is that the paths
aren't normalized.


joakim-brannstrom:
Nice, this solved the problem

PR #834, opened 2019-03-28
dextool: fix xml.d to store correct flags in _log.xml-file

peatersson:
Dextool seems to don't care if the usage is `dextool cpptestdouble [args]` or `dextool-cpptestdouble [args]` when storing the args in the _logs.xml-file.

Because of this, removing 1 flag (i.e. /path/to/dextool/dextool-cpptestdouble) is enough instead of 2 like before.

Will add tests for this, currently working on them.


joakim-brannstrom:
The files could use an dfmt formatting.
```sh
cd dextool_root
# because the .editorconfig is read by dfmt
dub fetch dfmt
dub run dfmt -- -i <file>
```

test/source/dextool_test/utils.d
```diff
@@ -575,3 +576,218 @@ void dirContentCopy(string src, string dst) {
             setAttributes(dst_f, attrs | S_IXUSR);
     }
 }
+
+// Check if commands are logged or not in the logfile (controlled with 'expected'-parameter)
+auto checkCommandsInLogFile(string[] commands, string file, bool expected) {
```
joakim-brannstrom: This can be replaced by the utilities moved from mutate?

peatersson: Yes, the reason I left the function was that it specified what you were looking for (commands in a logfile). But fair enough, it might be unnecessary. I'll remove it.

joakim-brannstrom:
The PR seems ready to me. Are you OK with merging it now?

peatersson:
Yes, that is perfectly fine for me.

joakim-brannstrom:
good job, thank you.

PR #871, opened 2019-07-03
mutant schemata

peatersson:
*** Work in progress - Do not merge ***

* Added lib/mutantschemata to Dextool. Two separate parts of mutantschemata
1. source/mutantschemata - D-side of the api. Is intended to work as an api to cpp code for inserting, selecting, deleting etc in the mutation database by utilizing miniorm.
2. cpp_source - Cpp-side of the api along with the implementation of schemata (currently placed in rewrite.hpp). "Links" external functions in the api together with the D-side of the code along with utility functionality.

* Changed compileflags for Dextool (impact of this needs to be further investigated)
-cppflags -> -cxxflags

* Has modified plugin/mutate/frontend/frontend.d to call mutantschemata in the analyze phase. This needs to be removed in order to not impact original functionality of Dextool Mutate.


cmake/Modules/FindLibClang.cmake
```diff
@@ -46,8 +46,8 @@ execute_process(COMMAND ${LLVM_CMD} version
     RESULT_VARIABLE llvm_config_VERSION_status
     OUTPUT_STRIP_TRAILING_WHITESPACE)
 
-execute_process(COMMAND ${LLVM_CMD} cpp-flags
-    OUTPUT_VARIABLE llvm_config_CPPFLAGS
+execute_process(COMMAND ${LLVM_CMD} cxx-flags
+    OUTPUT_VARIABLE llvm_config_CXXFLAGS
```
joakim-brannstrom: why is it necessary to rename it from cpp to cxx?
this should only have an effect on the variable name and nothing else thus no change in the behavior

cmake/introspect_llvm.d
```diff
@@ -109,11 +109,11 @@ string llvmVersion() {
...
 }
 
-string llvmCppFlags() {
-    const flags = execute([llvmCmd, "--cppflags"]);
+string llvmCxxFlags() {
+    const flags = execute([llvmCmd, "--cxxflags"]);
 
     // -std=c++0x is required to run on travis.
-    return flags.output.strip ~ " -std=c++0x -fno-exceptions -fno-rtti";
+    return flags.output.strip ~ " -std=c++0x -finline-functions -fno-strict-aliasing -fno-exceptions -fno-rtti";
```
joakim-brannstrom: Should be OK to use cxxflags and probably what should be done because the c++ code that exists is mainly to interact with clang.

But remove `-std=c++0x`. Why? As the comment states it where to make it work on travis. Shouldn't be an issue anymore.
Remove `-finline-functions`. Why? Unnecessary.
Remove `-fno-strict-aliasing`. Why? If it doesn't work without this flag then it is highly likely a bug in the code. The flag is a workaround.
Remove `-fno-exceptions` and `-fno-rtti`. Let llvm-config decide if these flags (via --cxxflags) should be used or not.

libs/mutantschemata/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+    -I${CMAKE_SOURCE_DIR}/plugin/mutate/source
+    -I${CMAKE_SOURCE_DIR}/vendor/d2sqlite3/source
+    -I${CMAKE_SOURCE_DIR}/vendor/sumtype/src
+    -I${CMAKE_SOURCE_DIR}/vendor/miniorm/source
+    -I${CMAKE_SOURCE_DIR}/dextool_clang_extensions/source
+    -I${CMAKE_SOURCE_DIR}/vendor/cachetools/source
+    -I${CMAKE_SOURCE_DIR}/vendor/arsd"
+)
+
+set(deps "dextool_d2sqlite3;dextool_sumtype;dextool_miniorm;dextool_mutantschemata_cpp_source")
```
joakim-brannstrom: move the `dextool_mutantschemata_cpp_source` to the beginning of this dependency list. Otherwise you may end up with hard to find linking errors.

libs/mutantschemata/cpp_source/type.hpp
```diff
@@ -0,0 +1,42 @@
...
+*/
+#ifndef TYPE_CPP
+#define TYPE_CPP
+
+#include "cpp_string.hpp"
+
+namespace CppType {
+
+struct SourceLoc {
+    unsigned line;
```
joakim-brannstrom: Be precise about the size of the variables by using types from `cstdint`. This is e.g. platform dependent on what size this is.

libs/mutantschemata/source/mutantschemata/api.d
```diff
@@ -0,0 +1,151 @@
...
+    string code;
+}
+SchemataFileString convert(SchemataFile sf) {
+    return SchemataFileString(cppToD!CppStr(sf.fpath),
+                                sf.mutants,
+                                cppToD!CppStr(sf.code));
+}
+
+// D class, connection to C++ code in /cpp_source
+class SchemataApi: SchemataApiCpp {
```
joakim-brannstrom: I think it is enough if you add `extern(C++)` here. Should thus not need to put it on every method.

libs/mutantschemata/source/mutantschemata/externals.d
```diff
@@ -0,0 +1,81 @@
...
+    }
+    extern (C++) struct CppStr {
+        void* cppStr;
+
+        const(void)* ptr();
+        int length();
+        void destroy();
+        void put(char);
+    }
+    extern (C++) CppBytes getBytes();
```
joakim-brannstrom: It is possible to write `extern(C++):`. Then everything statement after it is tagged as being `c++`.

peatersson:
Latest 3 commits where messy in terms of ordering, i am sorry for that. I blame the time of day.

I separated the db-handling in the mutantschemata-api. It helps both readability and helped with cleaning up the code (the comment about making all functions in the class extern).

I created a flag for the 'mutate analyze' part of Dextool Mutate so that by typing "--schemata yes", you can chose to execute mutantschemata instead of the original implementation. You don't actually need to type "yes" as a flag, you can use anything as the implementation is now.

I also cleaned up some code, removed tests in api.cpp and fixed a few other PR-comments.

libs/mutantschemata/cpp_source/api.cpp
```diff
@@ -14,51 +14,14 @@ in order to call D code and insert/select mutants from db obtained from Dextool
...
-
-  char cstr[cs.cppStr->size()+1];
-  strcpy(cstr, cs.cppStr->c_str());
-
-  setupClang(cstr, ".", ".");
+    std::cout << "here is: " << *cs.cppStr << std::endl;
+    //llvm::errs() << "Usage: rewritersample <file,otherfile,...> includeDir workingDir\n";
+    // return 1;
+    char cstr[cs.cppStr->size()+1];
+    strcpy(cstr, cs.cppStr->c_str());
```
joakim-brannstrom: Note that this is dangerous because cstr is stack allocated and a pointer to it escapes this function to setupClang. In this case it isn't a problem because setupClang copy the data to an internal stringstream.

`char cstr[cs.cppstr->size()+1]` will call [alloca](http://man7.org/linux/man-pages/man3/alloca.3.html) under the hood (see LLVM IR).

libs/mutantschemata/cpp_source/api.cpp
```diff
@@ -14,51 +14,14 @@ in order to call D code and insert/select mutants from db obtained from Dextool
...
 
+    setupClang(cstr, ".", ".");
```
joakim-brannstrom: Tip: to improve the safety of the function setupClang (make it easy to use correctly without data corruption) I recommend changing the paramters to `const std::string&`.

libs/mutantschemata/cpp_source/type.hpp
```diff
@@ -17,12 +17,12 @@ C++ types for easier communication between C++ and D code.
...
 struct SourceLoc {
-    unsigned line;
-    unsigned column;
+    unsigned int line;
+    unsigned int column;
 };
 struct Offset {
-    unsigned begin;
-    unsigned end;
+    unsigned int begin;
```
joakim-brannstrom: Tip: use `uint64_t` from `cstdint` here to ensure that it is always a 64bit unsigned int.
Then on the D side use `ulong`.
Then it will work on all platforms (32 and 64 bit).

See [C and D data type compatibility](https://dlang.org/spec/interfaceToC.html#data_type_compat).

libs/mutantschemata/cpp_source/api.cpp
peatersson: Sten will have to look into this since it affects rewrite.cpp (Mutant schemata implementation).

plugin/mutate/CMakeLists.txt
```diff
@@ -5,7 +5,25 @@ list(APPEND SRC_FILES ${CMAKE_SOURCE_DIR}/plugin/source/dextool/plugin/main/stan
 
 set(EXE_NAME ${DEXTOOL_MAIN_EXE}-mutate)
 
-set(flags "-I${CMAKE_CURRENT_LIST_DIR}/source -J${CMAKE_CURRENT_LIST_DIR}/view -I${CMAKE_SOURCE_DIR}/source -I${CMAKE_SOURCE_DIR}/dsrcgen/source -I${CMAKE_SOURCE_DIR}/clang -I${CMAKE_SOURCE_DIR}/libclang -I${CMAKE_SOURCE_DIR}/plugin/source -I${CMAKE_SOURCE_DIR}/dextool_clang_extensions/source -I${CMAKE_SOURCE_DIR}/vendor/d2sqlite3/source -J${CMAKE_SOURCE_DIR}/clang/resources -version=SqliteEnableColumnMetadata -version=SqliteEnableUnlockNotify -I${CMAKE_SOURCE_DIR}/vendor/taggedalgebraic/source -I${CMAKE_SOURCE_DIR}/vendor/colorlog/source -I${CMAKE_SOURCE_DIR}/vendor/toml/src -I${CMAKE_SOURCE_DIR}/vendor/arsd -I${CMAKE_SOURCE_DIR}/vendor/cachetools/source -I${CMAKE_SOURCE_DIR}/vendor/sumtype/src -I${CMAKE_SOURCE_DIR}/vendor/miniorm/source -I${CMAKE_SOURCE_DIR}/vendor/blob_model/source -I${CMAKE_SOURCE_DIR}/libs/mutantschemata/source")
+set(flags "-I${CMAKE_SOURCE_DIR}/libs/cpptooling/source -J${CMAKE_CURRENT_LIST_DIR}/view
+    -I${CMAKE_SOURCE_DIR}/source
+    -I${CMAKE_SOURCE_DIR}/libs/dextool/source
```
joakim-brannstrom: aye the restructured source code location. hopefully more logical now :)

peatersson: Exactly :)
I think I missed CMakeLists for schemata tho, so I pushed broken code :(

Will fix in 5

peatersson:
I don't know why the last azure-check failed, haven't seen that error before. I am leaving it up to Sten to fix if we want to pass all of them.

joakim-brannstrom:
> I don't know why the last azure-check failed, haven't seen that error before. I am leaving it up to Sten to fix if we want to pass all of them.

I think it is missing linking to `LLVMPolly.so`. I found the missing symbol in that lib by running this:
```
for F in `ls /usr/lib/llvm-8/lib/*.so`; do echo $F; readelf -Ws $F | grep -i countTrailingOnesSlowCase; done
```

libs/mutantschemata/cpp_source/rewrite.hpp
```diff
@@ -24,23 +24,171 @@
 #define MS_CLANG_REWRITE
 
 
-// global vars
+// defines to control what happens \\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\|
+//#define onlyAddSub
+#define debugDuplicateMutantPrevention
+// config mutant schemata (not really implemented)
+bool negateBinaryOperators = true;
```
joakim-brannstrom: I would recommend moving all globals to a "context" for the schemata. This enable such things as multithreading the schemata generation, easier reset (remember the case where an application isn't possible to inject all mutants in) etc.

libs/mutantschemata/cpp_source/rewrite.hpp
```diff
@@ -24,23 +24,171 @@
...
+class MutantRewriter: public clang::Rewriter {
+public:
+    bool InsertText(const clang::FileEntry *fE, unsigned StartOffs, StringRef Str, bool InsertAfter = true, bool indentNewLines = false) {
+        clang::FileID FID = this->getSourceMgr().translateFile(fE);
+        if (!FID.getHashValue()) {
+            FID = this->getSourceMgr().createFileID(fE, clang::SourceLocation(), clang::SrcMgr::CharacteristicKind::C_User);
+        }
+
+        llvm::SmallString<128> indentedStr;
+        if (indentNewLines && Str.find('\n') != StringRef::npos) {
```
joakim-brannstrom: suggestion: Even if it looks nice to generate correctly indented code it may complicate the source code.
I would recommend reducing the "formatting" to the bare minimum. If the user need it to look pretty they can use e.g. clang-format.
(I am assuming that indentNewLines mean that the schemata do the indentation)

cmake/Modules/FindLibClang.cmake
JKRT: Without knowing the context I might add that CPP and CXX is usually not the same thing. Usually CPP stands for the C Pre Processor and consists of another set of flags

libs/mutantschemata/cpp_source/cpp_string.cpp
```diff
@@ -0,0 +1,51 @@
...
+int CppStr::length() {
+    return cppStr->size();
+}
+void CppStr::destroy() {
+    delete cppStr;
+}
+void CppStr::put(char c){
+    cppStr->append(ONE_CHARACTER, c);
+}
+CppBytes getBytes(const char* text, int length) {
```
JKRT: Length should be const here

libs/mutantschemata/cpp_source/rewrite.hpp
```diff
@@ -0,0 +1,794 @@
...
+#include "clang/Tooling/CommonOptionsParser.h"
+#include "clang/Tooling/Tooling.h"
+#include "clang/Rewrite/Core/Rewriter.h"
+
+#ifndef MS_CLANG_REWRITE
+#define MS_CLANG_REWRITE
+
+
+// defines to control what happens \\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\|
+// config mutant schemata
```
JKRT: It seems that these booleans are used as flags. A more maintainable approach would be to assign them via macros passed via the -D flag when building dextool, then they could also be tested in separation

libs/mutantschemata/cpp_source/rewrite.hpp
```diff
@@ -0,0 +1,794 @@
...
+                    continue;
+            }
+            // insert the singleton
+            createAndStoreActualMutant(binOp, newExpr, endBracket);
+        }
+        
+    }
+    
+    void createAndStoreActualMutant(const clang::BinaryOperator *binOp, const std::string &newExpr, const std::string &endBracket) {
+
```
JKRT: Why big empty whitespace here?

libs/mutantschemata/cpp_source/rewrite.hpp
```diff
@@ -0,0 +1,794 @@
...
+        }
+    }
+    
+    /**
+     * Check if we want to mutate this file.
+     * (if it's a sourceFile and we haven't yet visited it)
+     */
+    bool isfileToMutate(clang::FullSourceLoc FullLocation) {
+        const clang::FileEntry *fE = SourceManager->getFileEntryForID(SourceManager->getFileID(FullLocation));
+        if (fE) {
```
JKRT: The indention of these ifblocks does not follow the rest of the file where the indention is 4 spaces

libs/mutantschemata/cpp_source/rewrite.hpp
```diff
@@ -0,0 +1,794 @@
...
+    
+    
+public:
+    explicit MutatingVisitor(clang::CompilerInstance *CI): astContext(&(CI->getASTContext())), pp(astContext->getLangOpts()) {
+        SourceManager = &astContext->getSourceManager();
+    }
+    
+    virtual bool VisitStmt(clang::Stmt *s) {
+        clang::FullSourceLoc FullLocation = astContext->getFullLoc(s->getLocStart());
+        if (!isfileToMutate(FullLocation)) {
```
JKRT: This if-statement is redundant you can just return the condition directly and take the boolean from it

JKRT: Sorry. The indention confused me. However, it seems that this function never returns false? When should it do that

libs/mutantschemata/cpp_source/rewrite.hpp
```diff
@@ -0,0 +1,794 @@
...
+                break;
+            default:
+                break;
+        }
+    }
+    
+    
+    void mutateBinaryLCR(clang::BinaryOperator *binOp) {
+        switch (binOp->getOpcode()) {
+                // Logical operators
```
JKRT: Instead of these comments, maybe write methods with names like handle_logical_operators and call them with suitable arguments

peatersson:
Latest commit was tested with a small example project. Both analyze and test-phase seems to work as intended. The only work needed was to manually add the Translation Unit constructor for the files mutated. This should however be fixed soon.

Setting environment variable is now done on the cpp-side as well, which worked just as intended.

libs/mutantschemata/cpp_source/api.cpp
```diff
@@ -48,19 +49,8 @@ void runSchemataCpp(SchemataApiCpp *sac, CppString::CppStr cs, CppString::CppStr
...
-    sm.offset = {0,3};
-    sm.status = 0;
-    sac->apiBuildMutant();
-    sac->apiInsertSchemataMutant(sm);
-    sm.mut_id = 365;
-    sm.status = 3;
-    sac->apiInsertSchemataMutant(sm);
-    */
+int setEnvironmentVariable(CppString::CppStr variable, CppString::CppStr value){
+  return setenv(variable.cppStr->c_str(), value.cppStr->c_str(), 1);
```
joakim-brannstrom: note: If the D side is driving the mutated binary then it is probably the call to std.process.spawnProcess that you want to inject the environment variables through.

joakim-brannstrom:
> Latest commit was tested with a small example project. Both analyze and test-phase seems to work as intended. The only work needed was to manually add the Translation Unit constructor for the files mutated. This should however be fixed soon.
> 
> Setting environment variable is now done on the cpp-side as well, which worked just as intended.

Aah ok. Can be improved later on. Make it work for now :)

peatersson:
Last commit also included some cleanup in terms of removing the use of templates (it was not needed) and unused functions.

I also added a new type for what is actually inserted into the DB, since inserting the CppStr seemed to be problematic. The solution is quite ugly, but it works for now.

peatersson:
Hmm, I am not sure. I just fixed it quick when i realized that the result in the DB were wrong. 

The way it is now seems reasonable, I think. When I do initialize (basically creating and setting the environment variable) I start with setting it to zero, which would mean that no mutants are executed if i run the test suite.

libs/mutantschemata/source/mutantschemata/type.d
```diff
@@ -0,0 +1,69 @@
...
+    Duration runtime;
+}
+struct DSchemataMutant {
+  ulong id;
+  ulong mut_id;
+  SourceLoc sourceLocation;
+  Offset offset;
+  ulong status;
+  string inject;
+  AbsolutePath filePath;
```
peatersson: @joakim-brannstrom 
This is where I feel you could help me with creating the checksums. The above struct is what we have in the database after analyze- and test-phase. You showed me some ideas before, but I imagine you would solve this both better and faster than me.

joakim-brannstrom: let me study it over the weekend

libs/mutantschemata/cpp_source/rewrite.hpp
```diff
@@ -268,7 +275,7 @@ void writeChangedFiles(bool inPlace) {
         for (std::map<clang::FileID, clang::RewriteBuffer>::iterator I = rewriter.buffer_begin(), E = rewriter.buffer_end(); I != E; ++I) {
             const clang::FileEntry *file = rewriter.getSourceMgr().getFileEntryForID(I->first);
             if (file) {
-                if (file->isValid()) {
+	      if (file->isValid()) {
```
peatersson: My emacs is completely bonkers with the indentations. Certainly something we solve later with format-checkers.

libs/mutantschemata/cpp_source/rewrite.hpp
```diff
@@ -346,11 +346,21 @@ class MutatingVisitor: public clang::RecursiveASTVisitor<MutatingVisitor> {
         assert(sema && "no Sema");
         sema->getDiagnostics().setErrorLimit(0);
         
+        bool rem_ok = true;
+        // TODO dirty hack to ensure that the rem operator does not cause compilation faults.
```
joakim-brannstrom: hehe "fulhack" ;-)

peatersson: kludge ;)

peatersson:
Mutant Schemata is now integrated to the extent that you can run it via Dextool. This way, error handling, help when analyzing and the config-file can be used. The integration does not affect the original version of Dextool Mutate since it is "hidden" behind flags.

For now, Mutant Schemata is only implemented for the analyze/generate mutant phase and the testing phase. The supported operators are ROR, AOR and LCR.

To analyze and generate mutants with Schemata instead of original version:
`dextool mutate analyze --schemata on`
( _on_ can be changed to pretty much anything. Dextool parses the input and checks that it is not empty.)

To test the mutants with Schemata instead of original version:
`dextool mutate test --schemata on`

As an example, Schemata has been tested on Googletest in _plugin/mutate/examples/gtest_. The "manual" touch needed in order to run Schemata on it is that the _-Werror_-flag needs to be removed from the compilation (one way is to just remove it from _googletest/googletest/cmake/internal_utils.cmake_)

**NOTE:** The current implementation of Schemata rewrites your files when injecting all the mutants. Reverting back to before the insertion of mutants needs to be done manually.

Issue #867, opened 2019-06-22
html: mutant killed by test case


joakim-brannstrom:
Show what test case killed a mutant.

joakim-brannstrom:
When inspecting a surviving mutant in the html code view, the user would like a convenient way to find test cases that killed a “near” mutant. For example if ‘a || b’ -> ‘false’ was killed, but ‘a || b’ -> ‘true’ survived, the user would like to know which test case(s) that killed the first mutation, so that they can use it to assess the surviving mutant (and perhaps extend the test suite to kill the second mutant).
Example suggestion: Add some ‘killed by test case(s): … ’ information box that shows up when the user is hovering the mouse over a mutation point that is killed.

christoffer-nylen:
In a formal verification process, an example use case could look something like this:

> ## How to handle surviving mutants?
>![coverage3](https://user-images.githubusercontent.com/5559216/60397704-ffedb180-9b4f-11e9-83af-84e0cdfe2f74.png)
>
>For high criticality software, suviving mutants of type SDL, DCR, LCR, LCRb must be resolved in one of the following ways:
>1. Remove code
>2. Correct tests
>3. Change/add requirements + tests
>4. Analysis and justification for e.g. defensive programming

**Problem:** For approach **2** and **3** the user would like to find out *how to kill a surviving mutant*.

**Solution:** Present the test cases that killed a “near” mutant, as suggested in previous comments.

Issue #866, opened 2019-06-22
html: actionable information


joakim-brannstrom:
Summarize the most important and actionable information as one report page in the html output.

What is actionable information? That is a good questions....

PR #879, opened 2019-07-12
HTML-report improvements: info line, mutant filters general improvements to usability and visual aspects

EricLindskog:


EricLindskog:
This also contains the d3.min package (v4) as it is required for the treemap.

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -145,11 +145,15 @@ struct FileIndex {
 
     override void endFileEvent(ref Database db) @trusted {
         import std.algorithm : max, each, map, min, canFind;
+        import std.algorithm.sorting : sort;
```
joakim-brannstrom: merge with 147. It should be possible to add `, sort` after canFind.

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -176,6 +180,21 @@ struct FileIndex {
         line.addClass("loc");
 
         line.addChild("span", "1:").addClass("line_nr");
+        auto mut_data = "var g_muts_data = {};\n";
+        mut_data ~= "g_muts_data[-1] = {'kind' : null, 'status' : null, 'testCases' : null, 'orgText' : null, 'mutText' : null, 'meta' : null};\n";
+        alias sort_tcs_on_kills = (a, b) => db.getTestCaseInfo(a, kinds)
```
joakim-brannstrom: This leads to two database queries. A specialized query that returns the test cases sorted is more efficient.
I will make a note of it as a TODO to fix in the future.

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -176,6 +180,21 @@ struct FileIndex {
...
+            .killedMutants < db.getTestCaseInfo(b, kinds).killedMutants;
+
+        //Finds the index of a kind in a list of kinds
+        auto find_kind_index = function(const Mutation.Kind a, const Mutation.Kind[] lkinds) {
```
joakim-brannstrom: Change it to a normal function definition as:
```
static auto findKindIndex(const Mutation.Kind a, const Mutation.Kind[] lkinds) {....}
```

joakim-brannstrom: I would recommend changing the index lookup to include all members of Mutation.Kind.
Then the use of this function could be replaced with a simple conversion to ubyte.
Pseudo-code:
```
addChild(new RawSource(ctx.doc, format("const g_mut_kind_map = [%('%s',%)'];", 
[EnumMembers!(Mutation.Kind))));
....
// find_kind_Index replaced with this
m.mut.kind.to!ubyte
```

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -205,48 +223,64 @@ struct FileIndex {
...
                         addClass("mutant");
                         addClass(s.tok.toName);
                         setAttribute("id", m.id.to!string);
-                        setAttribute("onmouseenter", fly);
-                        setAttribute("onmouseleave", fly);
                     }
                     d0.addChild("a").setAttribute("href", "#" ~ m.id.to!string);
+                    auto testCases = db.getTestCases(m.id);
+                    testCases.sort!(sort_tcs_on_kills);
+                    if (testCases.length)
```
joakim-brannstrom: wrap the branches in `{}`

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -205,48 +223,64 @@ struct FileIndex {
...
                     with (d0.addChild("span", m.mutation)) {
                         addClass("mutant");
                         addClass(s.tok.toName);
                         setAttribute("id", m.id.to!string);
-                        setAttribute("onmouseenter", fly);
-                        setAttribute("onmouseleave", fly);
                     }
                     d0.addChild("a").setAttribute("href", "#" ~ m.id.to!string);
+                    auto testCases = db.getTestCases(m.id);
+                    testCases.sort!(sort_tcs_on_kills);
```
joakim-brannstrom: Mmm this will result in many database queries. With the "fix" of a specialized DB query the testCases will already be sorted.
future improvement :)

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -287,13 +322,22 @@ struct FileIndex {
             File(fname, "w").write(fn());
         }
 
+        void addDataSource(Fn)(Fn fn, string name, string ext) {
+            import std.functional : unaryFun;
```
joakim-brannstrom: Unused import, remove.

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/page_tree_map.d
```diff
@@ -0,0 +1,68 @@
...
+@trusted auto makeTreeMapJSON(FileIndex[] files) {
+    import std.stdio : writeln;
+    import std.array : array, split;
+    import std.range : enumerate;
+    import std.json : JSONValue;
+
+    JSONValue root = ["name" : "root"];
+    auto curr = root;
+    foreach (f; files) {
+        auto path = f.display.split("/");
```
joakim-brannstrom: Use [pathSplitter](https://dlang.org/phobos/std_path.html#pathSplitter)

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/page_tree_map.d
```diff
@@ -0,0 +1,68 @@
...
+                    }
+                }
+                if (exists)
+                    continue;
+            }
+            JSONValue tmp = ["name" : seg];
+
+            if (i + 1 == path.length) {
+                tmp.object["locs"] = JSONValue(f.totalMutants);
+                if (f.totalMutants == 0)
```
joakim-brannstrom: Use `{}` for the branches. The second one is "somewhat long".

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/page_tree_map.d
```diff
@@ -0,0 +1,68 @@
...
+    import std.array : array, split;
+    import std.range : enumerate;
+    import std.json : JSONValue;
+
+    JSONValue root = ["name" : "root"];
+    auto curr = root;
+    foreach (f; files) {
+        auto path = f.display.split("/");
+
+        foreach (i, seg; path.enumerate) {
```
joakim-brannstrom: This is a bit hard to understand. I think my problem is that the code is both dependent on adding data to JSONValue and looking what has been added mixed in with an uneven control flow.
It would be easier, I think, if the first part transformed the data to an intermediary structure as "simple and straight" as possible. Trying to keep the control flow continuous.
Then the intermediary is converted to json.

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
EricLindskog: This function was a result of the old mutant filter, which filtered on the enum "Mutant" and not "MutantKind". This allowed us to only add the mutants actually used in the testing. This function ensured that we got the correct index in the potentially shorter list. This can be removed.

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/page_tree_map.d
```diff
@@ -1,60 +1,89 @@
...
-                    if (child["name"].str == seg) {// Segment already added to JSON
-                        parent = child; // Update parent
-                        continue segments; // Go to next segment
-                    }
-                }
-            } 
-            else { // Parent has no children -> init list
-                parent.object["children"] = JSONValue[].init;
+        auto parent = iRoot;
+        foreach (seg; path) {
```
EricLindskog: The intermediary class turned out to be a lot easier to manipulate, resulting in a cleaner "makeTreeMapJSON" function.

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/page_tree_map.d
```diff
@@ -1,60 +1,89 @@
...
-            parent.object["score"] = JSONValue(
-                    cast(double) f.killedMutants / cast(double) f.totalMutants);
+            parent = parent[seg];
         }
-        parent = root; //Finished with file -> set parent to root.
+        parent.locs = f.totalMutants;
+        parent.score = cast(double) f.killedMutants / cast(double) f.totalMutants;
     }
-    return root.toPrettyString;
+    return iRoot.toJSONValue().toPrettyString();
```
EricLindskog: All of the conversion to JSON is simply hidden behind this function call

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/page_tree_map.d
```diff
@@ -0,0 +1,100 @@
...
+        return children[i];
+    }
+
+    JSONLike opIndexAssign(JSONLike value, string i) pure nothrow {
+        children[i] = value;
+        return value;
+    }
+
+    bool opBinaryRight(string op)(string a) {
+        if (op == "in") {
```
joakim-brannstrom: I would recommend using a static if to turn any non-supported operator into a compilation error.
Example:
```
static if (op == "in") { ...}
else static assert(0, "Operator " ~ op ~ " not implmeneted");

File: plugin/mutate/source/dextool/plugin/mutate/backend/report/html/page_tree_map.d [hunk 13]
```diff
@@ -0,0 +1,100 @@
...
+        parent.locs = f.totalMutants;
+        if (f.totalMutants == 0)
+            parent.score = 1.0;
+        else
+            parent.score = cast(double) f.killedMutants / cast(double) f.totalMutants;
+    }
+    return iRoot.toJSONValue().toPrettyString();
+}
+
+auto makeTreeMapPage(FileIndex[] files) {
```
joakim-brannstrom: I would recommend re-arranging the source code so the public symbols are exposed at the top of the file. Then it is easy to add a blanket `private:` where the internal symbols start.
I think, in this case, it is only this function that is meant to be public so a "private:" after this function should do it?

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/page_tree_map.d
```diff
@@ -0,0 +1,100 @@
...
+            if (a in children)
+                return true;
+            return false;
+        }
+        return false;
+    }
+
+}
+
+@trusted auto makeTreeMapJSON(FileIndex[] files) {
```
joakim-brannstrom: Is this attribute needed? I think everything is actually `@safe` in this file.
Recommended reading: [memory safety in D](https://dlang.org/spec/memory-safe-d.html)

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/page_tree_map.d
```diff
@@ -0,0 +1,100 @@
...
+
+/* 
+* A JSON-like object that is more easily manipulated than JSONValue.
+* Has operator overloads for "index", "index assign" and "in" allowing the children list to be private
+* and cleaner construction of the object. 
+*/
+class JSONLike {
+    string name;
+    private JSONLike[string] children;
+    long locs = -1;
```
joakim-brannstrom: Tip: Instead of a "magic value" that signals a behaviour you can use `Nullable!long`. It default initializes to "null".
Then it is possible at line 30 to check if it contains a value via `.isNull` and if so access the value via `.get`.

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/page_tree_map.d
```diff
@@ -0,0 +1,100 @@
...
+    double score = -1;
+
+    JSONValue toJSONValue() {
+        JSONValue root = ["name" : name];
+
+        if (children.length) {
```
joakim-brannstrom: I recommend using `.empty` from `std.array`. It makes the intent more clear.

joakim-brannstrom:
Good progress :)

joakim-brannstrom:
Thanks both of you for the contribution to dextool. This will be a significant improvement to the users understanding of the mutation testing result.

joakim-brannstrom:
Feature for #867

joakim-brannstrom:
The tree map view is partial for #866

PR #884, opened 2019-08-13
mutate: update requirements

joakim-brannstrom:


joakim-brannstrom:
@christoffer-nylen , I tried to capture what you wrote in #867 . Do you think this is good enough?

joakim-brannstrom:
Requirements for #867

PR #889, opened 2019-08-26
Fixed macOS linkerflags

peatersson:
Removed the flags for osSpecificLinkerFlags so that Dextool can be built on macOS.

### DIFF
--- a/cmake/introspect_llvm.d
+++ b/cmake/introspect_llvm.d
@@ -59,12 +59,12 @@ string[] osSpecificLinkerFlag() {
     version (OSX) {
         return null;
     } else {
-        return ["-Wl,--enable-new-dtags"];
+        return ["-Wl,--enable-new-dtags,-Wl,--no-as-needed"];
     }
 }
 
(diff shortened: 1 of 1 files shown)


cmake/introspect_llvm.d
```diff
@@ -59,12 +59,12 @@ string[] osSpecificLinkerFlag() {
     version (OSX) {
         return null;
     } else {
-        return ["-Wl,--enable-new-dtags"];
+        return ["-Wl,--enable-new-dtags,--no-as-needed"];
```
joakim-brannstrom: I would recommend splitting this in two for consistency.
```d
["-Wl,--enable-new-dtags", "-Wl,--no-as-needed"]
```

I have to admit that I didn't know that the linker supported multiple arguments via `-Wl`.....

joakim-brannstrom: I mean, the CI is green so obviously it do ;-)

joakim-brannstrom:
Thank you for the fix

Issue #920, opened 2019-09-18
Add creation of alias to game_tutorial example


peatersson:
Add the following to README.md in _plugin/mutate/examples/game_tutorial_:

```
Set an alias for dextool:
+```sh
+alias dextool="path/to/dextool/"
+```
+(**NOTE**: Change *path/to/dextool* to the correct path to were you previously installed it. For example: */home/myUser/dextool/build/bin/dextool*. This is needed when you want to run dextool in coming steps as well, we therefore create an alias.
```

peatersson:
Perfect! :)
It sure is a quality of life thing to do. It also reduces the amount of "NOTE: change path to .." in every commando presented.

PR #921, opened 2019-09-18
mutate: update readme of the example

joakim-brannstrom:
#920


joakim-brannstrom:
@peatersson is this as you where thinking?

Issue #918, opened 2019-09-16
mutate: missing sdl mutant for return statements in void functions


christoffer-nylen:
I have code that looks something like this:
```
122: void createStuff() {
...
138:   if (numCreated >= MAX)
139:     *result = INVALID_NUM;
140:     return;
141:   }
142:
143:   if (id == null){
144:     *result = INVALID_ID;
145:     return;
146:   }
147:
148:   *result = OK;
149: }
```

I used the `sdl` mutation operator and got the following html result:

Line 139 is red.
Line 140 is green.
Line 141 is green.

Line 144 is red.
Line 145 is green.
Line 146 is green.

Problem/bug:
* The list of test cases does not appear when selecting line 140 and 145.

Comment:
* I suspect line 140 and 145 is green due to function body deletion but single-line mutants is not generated for `return;` statements?

joakim-brannstrom:
solved

PR #924, opened 2019-09-23
mutate: delete return statement of void functions

joakim-brannstrom:
#918


joakim-brannstrom:
Still working on the test case

christoffer-nylen:
Awesome!! 👍

joakim-brannstrom:
The fix for the hidden AST node is to replace `accept` with `dispatch`. Accept visits the children while dispatch the current node. The result of this code change is that we visit the nodes in a if-statement two times.

PR #936, opened 2019-10-14
docker: bump llvm

joakim-brannstrom:


joakim-brannstrom:
I have changed llvm to version 8. This should help with the mutation schemata.

Issue #938, opened 2019-10-17
Dextool can't handle path with space in foldername


peatersson:
I got a tip that maybe highlighted a bug in Dextool. If I create a folder with spaces in the name, the analyze phase in Dextool Mutate is not able to compile or process the files correctly.

Output:
```
~/Projects/mutantninjas/test project  $: dextool mutate analyze

info: Using /Users/niklas/Projects/mutantninjas/test project/dextool_mutate.sqlite3
info: Upgrading database from 0
info: Reading compilation database:
/Users/niklas/Projects/mutantninjas/test project/build/compile_commands.json
info: Analyze and mutation of files will only be done on those inside this directory root
info:   User input: .
info:   Real path: /Users/niklas/Projects/mutantninjas/test project
info: Further restricted to those inside these paths
info:   User input: src/
info:   Real path: /Users/niklas/Projects/mutantninjas/test project/src
info: Analyzing '/Users/niklas/Projects/mutantninjas/test project/src/fibonacci_example.cpp'
info: Analyzing '/Users/niklas/Projects/mutantninjas/test project/tests/fibonacci_test.cpp'
error: /Users/niklas/Projects/mutantninjas/test project/tests/fibonacci_test.cpp:3:10: fatal error: 'fibonacci_example.cpp' file not found
error: Compile error...
```

In _fibonacci_test.cpp_, I include _fibonacci_example.cpp_, but it seems like Dextool can't handle the space in the folder name _test project_. The project itself compiles and run without problem.

I am not sure we even need to support this, since it seems a bit unconventional to name folders with spaces. Creating an issue for it though seemed like the way to go.

peatersson:
Cred to @JKRT for bringing this up.

joakim-brannstrom-work1:
It works "here".... I think it may have something to do with the cmake generate compile_commands.json

peatersson:
When I looked at the compile_commands.json generated with cmake, it added quation-marks (") around the paths. This is why it worked nomally when building and testing, right?

peatersson:
This is the compilation_database.json generated by cmake:
```
[
{
  "directory": "/home/niklas/Projects/mutantninjas/test project/build/src",
  "command": "/usr/bin/c++  -Dcmake_template_EXPORTS  -fPIC   -o CMakeFiles/cmake_template.dir/decrement_example.cpp.o -c \"/home/niklas/Projects/mutantninjas/test project/src/decrement_example.cpp\"",
[...]

PR #940, opened 2019-10-21
mutate: added so restrictDir is used in analyze-phase.

peatersson:
In [#938](https://github.com/joakim-brannstrom/dextool/issues/938), I noticed that the analyze-phase was not restricted to the correct path, as given in the configuration file. The addition in this PR is for handling that, and not the rest of the issue.

Since the FrontendValidateLoc was setup earlier in frontend.d, where the restrictDir was used, the only addition needed was to call the _shouldAnalyze_-function.

### DIFF
--- a/plugin/mutate/source/dextool/plugin/mutate/backend/analyze/package.d
+++ b/plugin/mutate/source/dextool/plugin/mutate/backend/analyze/package.d
@@ -112,6 +112,8 @@ struct Analyzer {
 
         if (analyzed_files.contains(checked_in_file))
             return;
+        if (!val_loc.shouldAnalyze(checked_in_file))
+            return;
 
         analyzed_files.add(checked_in_file);
 


joakim-brannstrom-work1:
This has to be a separate configuration parameter that goes into the `[compile_commands]` section that affects the range of files that are analyzed.

joakim-brannstrom-work1:
https://github.com/joakim-brannstrom/dextool/blob/ce80ab22128c6f6f97c7d4519327cf8531f7fc72/libs/dextool/source/dextool/compilation_db/package.d#L728

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -739,19 +761,25 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
...
+	    } else if (st == State.checkingForQuotation) {
+		path.put(" " ~ arg);
+		if (isEndingWithQuotation(arg)) {
+		  st = State.keep;
+		  putNormalizedAbsolute(path.data.strip('"'));
+		}
             } else if (excludeStartWith(arg, flag_filter)) {
                 st = State.skipIfNotFlag;
-            } else if (isCombinedIncludeFlag(arg)) {
+	    } else if (isSpacedInclude(arg)) {
```
JKRT: Indention is a bit wonky here

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -714,16 +727,25 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
...
         import std.array : appender, array;
         import std.range : ElementType;
 
         auto st = State.keep;
         auto rval = appender!(string[]);
         auto includes = appender!(string[]);
         auto compiler = Compiler(r.length == 0 ? null : r[0]);
+	auto path = appender!(string);
+
+	void putNormalizedAbsolute(string arg) {
```
JKRT: You can probably make arg const a tiny little micro optimisation.

How does D work, is const & a thing?

libs/dextool/source/dextool/compilation_db/package.d
peatersson: I know, I wrote it earlier that my formatchecker was acting up.

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -647,6 +647,8 @@ struct ParseFlags {
 ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
         const Compiler user_compiler = Compiler.init) @safe {
     import std.algorithm : among, map;
+    import std.algorithm.mutation: strip;
```
joakim-brannstrom: You can move the include to after map on line 649 :)

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -701,6 +703,17 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
         return 0 != flag.among("-D");
     }
 
+    /// Flags that is includes but contains spaces are wrapped in quotation marks.
+    static bool isSpacedInclude(string flag) @safe {
```
joakim-brannstrom: I think you need to do something like `isCombinedIncludeFlag` and `isFlagAndPath` because both these are valid arguments that we need to detect:
`-I"foo/bar"` and `-I "foo/bar`. Notice the space.... yes there exists build systems in the world that do it that way...

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -701,6 +703,17 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
...
+	return flag[0 .. 2] == "-I" && indexOf(flag, '"') == 2;
```
joakim-brannstrom: Have to check the length first before slicing. Otherwise if the string is ever less than 2 chars this will raise an exception.

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -701,6 +703,17 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
...
+    }
+
+    /// When we know we are building a path that is space separated,
+    /// the last index of the last string should be a quotation mark.
+    static bool isEndingWithQuotation(string flag) @safe {
+	return indexOf(flag, '"') == (flag.length-1);
```
joakim-brannstrom: Another way of doing it is `!flag.empty && flag[$-1] == '"'`. The `$` is the last index.

libs/dextool/source/dextool/compilation_db/package.d
joakim-brannstrom: `ref string` would be the equivalent.
But string is actually already an alias for `const(char)[]`.

joakim-brannstrom: make it a function by adding `static`. Then we know at compile time that it doesn't use any local variables (e.g. is a delegate).

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -739,19 +761,25 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
...
                 st = State.keep;
-                auto p = buildNormalizedPath(workdir, arg).absolutePath;
-                rval.put(p);
-                includes.put(p);
+                putNormalizedAbsolute(arg);
             } else if (st == State.priorityKeepNextArg) {
                 st = State.keep;
                 rval.put(arg);
+	    } else if (st == State.checkingForQuotation) {
+		path.put(" " ~ arg);
```
joakim-brannstrom: Change to
```d
path.put(" ");
path.put(arg);
```
It is more efficient because it avoid an allocation.

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -739,19 +761,25 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
...
+		if (isEndingWithQuotation(arg)) {
+		  st = State.keep;
+		  putNormalizedAbsolute(path.data.strip('"'));
```
joakim-brannstrom: The quotation mark should not be added to path. Internally and when executing external programs it isn't needed anymore. It is just "there" for e.g. bash expansion etc.

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -739,19 +761,25 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
...
-                auto p = buildNormalizedPath(workdir, arg).absolutePath;
-                rval.put(p);
-                includes.put(p);
+                putNormalizedAbsolute(arg);
             } else if (st == State.priorityKeepNextArg) {
                 st = State.keep;
                 rval.put(arg);
+	    } else if (st == State.checkingForQuotation) {
+		path.put(" " ~ arg);
+		if (isEndingWithQuotation(arg)) {
```
joakim-brannstrom: I think `path.clear` has to be called after it is used because it may be reused for another "string building".

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -714,16 +730,25 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
...
         import std.path : buildNormalizedPath, absolutePath;
-        import std.array : appender, array;
+        import std.array : Appender, appender, array, join;
         import std.range : ElementType;
 
         auto st = State.keep;
         auto rval = appender!(string[]);
         auto includes = appender!(string[]);
         auto compiler = Compiler(r.length == 0 ? null : r[0]);
+        Appender!(string[]) path;
```
joakim-brannstrom: The type should probably be `string` because the purpose is to go through an array of strings (`string[]`) to create the sum of them as a path.

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -739,19 +764,31 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
...
                 st = State.keep;
                 rval.put(arg);
+            } else if (st == State.checkingForQuotation) {
+                if (!path.data.empty) {
+                    path.put(" ");
+                }
+                path.put(arg);
+                if (isEndingWithQuotation(arg)) {
+                    st = State.keep;
+                    putNormalizedAbsolute(path.data.join.strip('"'));
```
joakim-brannstrom: By changing the type of `path` to string this can be made more efficient by changing it to `path.data[0 .. $-1]`

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -1180,3 +1216,29 @@ unittest {
 
     found[0].absoluteFile.baseName.shouldEqual("file3.cpp");
 }
+
+@("shall extract filepath from includes correctly when there is spaces in the path")
+unittest {
+    auto cmd = toCompileCommand("/home", "file.cpp", [
+            "-I", "dir with spaces/src"
```
joakim-brannstrom: It should probably be:
```d
"-I", `"dir with spaces/src"`
```

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -1180,3 +1216,29 @@ unittest {
...
+            ], AbsoluteCompileDbDirectory("/home"), null);
+    auto pargs = cmd.get.parseFlag(defaultCompilerFilter, Compiler.init);
+    pargs.cflags.shouldEqual(["-I", "/home/dir with spaces/src"]);
+    pargs.includes.shouldEqual(["/home/dir with spaces/src"]);
+}
+
+@("shall handle path with spaces, both as separate string and combined")
+unittest {
+    auto cmd = toCompileCommand("/project", "file.cpp", [
+            "-I", "separate dir/with space", "-Icombined dir/with space/src"
```
joakim-brannstrom: See my previous comment for the first argument.
The second one should probably be:
```d
`-I"combined dir/with space/src"`
```

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -714,16 +730,25 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
             skip,
             /// skip the next arg, if it is not a flag
             skipIfNotFlag,
+            /// use the next arg to create a complete path
+            checkingForQuotation,
```
joakim-brannstrom: `checkingForEndQuotation` ?

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -739,19 +764,31 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
...
+                    st = State.keep;
+                    putNormalizedAbsolute(path.data.join.strip('"'));
+                    path.clear();
+                }
             } else if (excludeStartWith(arg, flag_filter)) {
                 st = State.skipIfNotFlag;
+            } else if (isSpacedInclude(arg)) {
+                st = State.checkingForQuotation;
+                rval.put("-I");
+                if (!arg[2 .. $].empty) {
```
joakim-brannstrom: I recommend changing this to `arg.length >= 4`. It namely took me a while to "see" that it is safe to take a slice like this because `isSpacedInclude` is only true if the length is >= 3.

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -739,19 +764,31 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
...
+                    path.put(arg[2 .. $]);
```
joakim-brannstrom: This most probably should be `arg[3 .. $]` because arg is in this case `-I"foo bar`. This, I think, will be shown when the test case is adjusted to contain a quotation mark

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -1220,25 +1254,54 @@ unittest {
...
+            "/project/one space/lots of     space",
             ]);
     pargs.includes.shouldEqual([
-            "/project/separate dir/with space",
-            "/project/combined dir/with space/src"
+            "/project/one space/lots of     space",
+            "/project/one space/lots of     space",
+            "/project/one space/lots of     space",
+            "/project/one space/lots of     space"
             ]);
```
peatersson: These test cases are meant to test the example provided in #938. When printing the args in the filterPair-function, the following is the output:
```
-I"/home/niklas/Projects/mutant
ninjas/test
project/tests/../src"
```
where newlines are spaces in the path. I am not sure, however, that two of the cases can ever happen. For example:
```
`-I`, `\"one`, `space/lots`, `of     space\"`
```
Anyhow, the test is there, and the implementation handles it. Should we keep it "in case of"?

libs/dextool/source/dextool/compilation_db/package.d
```diff
@@ -714,16 +741,33 @@ ParseFlags parseFlag(CompileCommand cmd, const CompileCommandFilter flag_filter,
...
         auto rval = appender!(string[]);
         auto includes = appender!(string[]);
         auto compiler = Compiler(r.length == 0 ? null : r[0]);
+        string path;
+        const string EMPTY_STRING;
+
+        string removeBackslashesAndQuotes(string arg) {
+            import std.string : replace;
+
+            return arg.replace(`\"`, EMPTY_STRING).replace(`"`, EMPTY_STRING);
```
joakim-brannstrom: We have to assume that this may contain complex utf-8 strings because the underlying filesystem for do so.
This complicates the handling of the strings as they then have to be handled with care.
A utf-8 string contains code units, code points and grapheme. A grapheme is made up of one or more code points, and a code point is made up of one or more code units.
An iteration over a char[] as-is will iterate over the code points. There are no guarantee that a code point actually is a representable character. A code unit is just a piece of a character except for those rare cases where the input is ASCII.
A code point though is actually something representable. Same with a grapheme which can for example be a modifier for a character such as the dots above o in a ö. This fast becomes complicated.

Thus how to solve this conundrum?.
```d
return arg.byGrapheme.filter!(a => !a.among(Grapheme('\\'), Grapheme('"')).byCodePoint.text
```

libs/dextool/source/dextool/compilation_db/package.d
joakim-brannstrom: Correct, the case above can never happen.
it is either:
```d
[`-I`, `\\"one space/lots of     spaces\\"`]
```
or
```d
[`-I\\"one space/lots of     spaces\\"`]
```

PR #943, opened 2019-10-30
mutate: configurable max runtime during test phase

joakim-brannstrom:


peatersson:
It looks very nice. Updating requirements is very good, soon we will have a complete documentation. :)

PR #944, opened 2019-11-02
mutate: allow the user to exclude files from being analyzed for mutants

joakim-brannstrom:
This speeds up the analysis because the user can exclude those files
that aren't of interest to mutant such as external libraries.


joakim-brannstrom:
missing test case. I will add it later.

peatersson:
This is very nice! 

@Sten-Vercammen , this is useful for us in OpenCV.

PR #947, opened 2019-11-04
mutate: fix documentation of the changes to mutation score calculation

joakim-brannstrom-work1:


joakim-brannstrom-work1:
@christoffer-nylen , as you can see I changed the requirement to what it should have been from the start.

Issue #917, opened 2019-09-16
mutate: add customizable filter for unproductive mutants


christoffer-nylen:
Assume:
* There is a file `a.c` with 534 LoC.
* The file contains 37 invokations of a function `addLoggMessage(param, param)` that is used for debugging purposes, and there are 359 similar cases in the full project.
* The `addLoggMessage` usage (e.g., invokation order and contents of argument) is highly coupled to the `a.c` implementation.
* The correctness of the `addLoggMessage()` usage is not critical.

Scenario:
* The user would like to focus on critical/productive test scenarios.

Problem:
* In `a.c`, 37 invokations of `addLoggMessage(param, param)` gets reported as survived.
* The reported statements draw attention from critical problems.

Suggestion:
* Add some kind of filtering function so that the user can specify to ignore all invokations of `addLoggMessage()`.

joakim-brannstrom:
Yes this is coupled to how to improve NOMUT. Suppressing mutants.
I have had the idea to let the user specify uninteresting functions.

joakim-brannstrom:
For now what exist is to mark those lines with NOMUT

peatersson:
The idea I've started implementing, but put on hold, was that the user specified mutantID. 
Example:
`dextool mutate admin --suppress 52`

I realized that this quickly became problematic in terms of scale. A user cant be expected to write that commando for every single mutant, if there are 359 as in your example. 

I know I wouldn't.

Are we maybe thinking of an addition to the config-file? Something like:
```
# Specify which functions you would like to exclude from the mutation testing
# Note: this will affect all kinds of mutants surrounding those functions.
# For example: SDL won't remove the call to the function, or the function-body.
suppress-functions = ["addLoggMessage", "print", "randomFunctionToSuppress"]
```
And then we add the ability to run admin-mode and suppress those now added to the config?

joakim-brannstrom:
> The idea I've started implementing, but put on hold, was that the user specified mutantID.

Did you consider suppressing a whole line?

> Example:
> `dextool mutate admin --suppress 52`
> 
> I realized that this quickly became problematic in terms of scale. A user cant be expected to write that commando for every single mutant, if there are 359 as in your example.

I agree. Another problem is that they constantly change. The checksum is *more* stable.

> Are we maybe thinking of an addition to the config-file? Something like:
> 
> ```
> # Specify which functions you would like to exclude from the mutation testing
> # Note: this will affect all kinds of mutants surrounding those functions.
[...]

PR #953, opened 2019-11-11
Mutate: mark mutants in db via admin-command

peatersson:
See #917 for more discussion regarding the addition of this feature.

Summary:
- User can mark mutants with a status and rationale via admin command _--operation markMutant_.
- Dextool will notify when marked mutants are "lost" due to rerunning analyze
- User can view marked mutants with report command _--section marked_mutants_ (for now only available in plain-style)
- Feature is tested and documented in requirements


plugin/mutate/source/dextool/plugin/mutate/backend/database/type.d
```diff
@@ -225,3 +225,17 @@ struct MutantMetaData {
...
 }
+
+struct Rationale {
+  string time;
+  string rationale;
+}
+
+// Mutant manually marked in db
+struct MarkedMutant {
+    MutationId mutant_id;
```
joakim-brannstrom: This should be a mutation status id.
The mutationId will change each analyze thus it becomes invalid almost directly when used in practise.
It should also be marked as the primary key via the necessary UDA.

plugin/mutate/source/dextool/plugin/mutate/backend/database/type.d
```diff
@@ -225,3 +225,17 @@ struct MutantMetaData {
         attr.match!((NoMetadata a) {}, (NoMut a) => put(w, "nomut"));
     }
 }
+
+struct Rationale {
```
joakim-brannstrom: Good that a specific Rationale type is created. It gives more information to to the programmer and the type system compared to a plain `string`.

plugin/mutate/source/dextool/plugin/mutate/backend/database/type.d
```diff
@@ -225,3 +225,17 @@ struct MutantMetaData {
...
+  string time;
```
joakim-brannstrom: I think that this time should be part of the MarkMutant, not the rationale.

plugin/mutate/source/dextool/plugin/mutate/backend/database/type.d
```diff
@@ -225,3 +225,17 @@ struct MutantMetaData {
...
+  string rationale;
+}
+
+// Mutant manually marked in db
+struct MarkedMutant {
+    MutationId mutant_id;
+    Path path;
+    Mutation.Status from_status;
```
joakim-brannstrom: Why is a `from_status` needed?

plugin/mutate/source/dextool/plugin/mutate/backend/database/type.d
```diff
@@ -225,3 +225,17 @@ struct MutantMetaData {
...
     }
 }
+
+struct Rationale {
+  string time;
+  string rationale;
+}
+
+// Mutant manually marked in db
+struct MarkedMutant {
```
joakim-brannstrom: Missing the line and column?

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -524,6 +534,25 @@ struct Database {
...
+        // mark mutantStatus in mutationStatus-table
+        const s = format!"UPDATE %s SET status=%s WHERE id IN (SELECT st_id FROM %s WHERE id = %s)"(
+                mutationStatusTable, to_status.to!long, mutationTable, id.to!long);
+        auto stmt = db.prepare(s);
+        stmt.execute;
+
+        // add mutant to separate table for marked mutants
+        db.run(buildSchema!MarkedMutant); // TODO: can probably be done somewhere else
+        db.run(insert!MarkedMutant.insert, MarkedMutant(MutationId(id),
+                Path(getPath(MutationId(id))), from_status, to_status,
```
joakim-brannstrom: fetching the path should be done before this call because it can fail if the user provides an invalid ID

plugin/mutate/doc/design/options.md
```diff
@@ -75,6 +76,12 @@ The flags for the different commands will be listed in the following format:
     - *killedByCompiler* : Invalid mutants generated that caused the compilation of the project to fail.
     - *timeout* : Mutants that timed out during test suite execution.
 
+
+-   --id : Specify a specific mutant by MutationId.
```
joakim-brannstrom: MutationId is an internal, programming terminology? What does the html report say?

plugin/mutate/source/dextool/plugin/mutate/backend/database/type.d
```diff
@@ -225,3 +225,17 @@ struct MutantMetaData {
...
+    MutationId mutant_id;
+    Path path;
```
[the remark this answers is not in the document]
peatersson: My thought was that the information is then used when the analyze-phase is rerun in order to tell the user which marked mutants will become obsolete (not yet implemented).

Or are you meaning that I could use AbsolutePath, instead of Path?

plugin/mutate/source/dextool/plugin/mutate/backend/database/type.d
peatersson: It was just added for more information, i.e. letting the user know that "the mutant you marked last week was alive, and you marked it killed because of rationale" etc. 

But you are correct, it might not be needed.

plugin/mutate/source/dextool/plugin/mutate/backend/database/type.d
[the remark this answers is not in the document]
joakim-brannstrom: Path is good. I just didn't understand that it's was too be used as information. Path, line, column and annotation should be pretty good for the user

plugin/mutate/source/dextool/plugin/mutate/backend/database/type.d
joakim-brannstrom: I would recommend removing it if the following statement holds water :
A user will most likely use the mark mutant to test alive mutants. There are no other case that would be a useful scenario /correct use

peatersson: It doesn't? A user would most likely use markMutant to mark alive mutants killed without modifying the code. There are other useful uses as well. Marking "bad" mutants as either killed or killedByCompiler is completely valid. Even if it might be considered misuse and could be abused, it is up to the user?

I still agree though that it could be removed. So, I will remove it in coming commit. :)

plugin/mutate/source/dextool/plugin/mutate/backend/database/type.d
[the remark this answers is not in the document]
peatersson: Should we maybe add mutant kind?

plugin/mutate/source/dextool/plugin/mutate/backend/admin.d
```diff
@@ -77,13 +77,9 @@ nothrow:
         return this;
     }
 
-    auto mutant_id(ulong v) {
+    auto markMutantData(ulong v, string s) {
```
joakim-brannstrom: MutationId is a wrapper around a long thus v here should be `long`

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -411,6 +412,30 @@ struct MutantTimeoutCtx {
     State state;
 }
 
+import dextool.plugin.mutate.backend.database.type : Rationale;
+
+@TableName(markedMutantTable)
+@TablePrimaryKey("mutationStatusId")
+struct MarkedMutant {
+    @ColumnName("st_id")
+    ulong mutationStatusId;
```
joakim-brannstrom: `long` as per the previous comment

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -411,6 +412,30 @@ struct MutantTimeoutCtx {
     State state;
 }
 
+import dextool.plugin.mutate.backend.database.type : Rationale;
+
+@TableName(markedMutantTable)
+@TablePrimaryKey("mutationStatusId")
```
joakim-brannstrom: should probably be `st_id` because the name is changed by the `ColumnName` UDA further down

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -411,6 +412,30 @@ struct MutantTimeoutCtx {
...
+    @ColumnName("st_id")
+    ulong mutationStatusId;
+
+    uint line;
+
+    uint column;
+
+    string path;
+
+    @ColumnName("status")
```
joakim-brannstrom: Might as well change the name of `to_status` to status which would make this UDA redundant

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -411,6 +412,30 @@ struct MutantTimeoutCtx {
...
+    ulong to_status;
+
+    @ColumnName("time")
+    SysTime time;
+
+    @ColumnName("rationale")
```
joakim-brannstrom: You have remove this UDA. the name should be, automatically, the same as the field name

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -411,6 +412,30 @@ struct MutantTimeoutCtx {
...
+    uint line;
+
+    uint column;
+
+    string path;
+
+    @ColumnName("status")
+    ulong to_status;
+
+    @ColumnName("time")
```
joakim-brannstrom: same here, this should be possible to remove

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -480,7 +505,7 @@ void upgradeV0(ref Miniorm db) {
     enum tbl = makeUpgradeTable;
 
     db.run(buildSchema!(VersionTbl, RawSrcMetadata, FilesTbl, MutationPointTbl, MutationTbl, TestCaseKilledTbl,
-            AllTestCaseTbl, MutationStatusTbl, MutantTimeoutCtx, MutantTimeoutWorklist));
+            AllTestCaseTbl, MutationStatusTbl, MutantTimeoutCtx, MutantTimeoutWorklist, MarkedMutant));
```
joakim-brannstrom: good good

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -807,6 +832,12 @@ void upgradeV12(ref Miniorm db) {
     updateSchemaVersion(db, 13);
 }
 
+/// 2019-11-12
+void upgradeV13(ref Miniorm db) {
```
joakim-brannstrom: good

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -401,7 +401,7 @@ struct Database {
     }
     // Returns: the status of the mutant
     Nullable!Mutation.Status getMutationStatus(const MutationId id) @trusted {
-        auto stmt = db.prepare(format("SELECT status FROM %s WHERE id IN (SELECT st_id FROM %s WHERE id=:mut_id)",
+        auto stmt = db.prepare(format!"SELECT status FROM %s WHERE id IN (SELECT st_id FROM %s WHERE id=:mut_id)"(
```
joakim-brannstrom: i recommend moving the string to a line before this and assigned to an immutable variable.

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -515,6 +515,18 @@ struct Database {
         return app.data;
     }
 
+    /** Get SourceLoc for a specific mutation id.
+     */
+    SourceLoc getSourceLocation(MutationId id) @trusted {
+        auto stmt = db.prepare(format!"SELECT line, column FROM %s WHERE id IN (SELECT mp_id FROM %s WHERE id=:mut_id)"(
```
joakim-brannstrom: same here, I recommend moving the SQL to a line before this and assigned to an immutable variable. forces it to be computed at compile time and only once.

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -515,6 +515,18 @@ struct Database {
         return app.data;
     }
 
+    /** Get SourceLoc for a specific mutation id.
+     */
+    SourceLoc getSourceLocation(MutationId id) @trusted {
```
joakim-brannstrom: Nullable?

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -535,22 +547,21 @@ struct Database {
...
+    void markMutant(MutationId id, Mutation.Status to_status, string rationale) @trusted {
+        auto st_id = getMutationStatusId(id);
+        auto p = getPath(id);
 
-        // mark mutantStatus in mutationStatus-table
-        const s = format!"UPDATE %s SET status=%s WHERE id IN (SELECT st_id FROM %s WHERE id = %s)"(
-                mutationStatusTable, to_status.to!long, mutationTable, id.to!long);
-        auto stmt = db.prepare(s);
-        stmt.execute;
+        if (p.isNull)
```
joakim-brannstrom: I recommend considering how to handle this failing case.
How will the user know that an invalid mutation id was specified?

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -535,22 +547,21 @@ struct Database {
     }
 
     /** Set mutantStatus by using MutantId.
-	Can be used both for ignoring and "accepting" mutants.
      */
-    void markMutant(ulong id, Mutation.Status to_status, string rationale = "") @trusted {
-        Mutation.Status from_status = Mutation.Status(getMutationStatus(MutationId(id)));
+    void markMutant(MutationId id, Mutation.Status to_status, string rationale) @trusted {
+        auto st_id = getMutationStatusId(id);
```
joakim-brannstrom: i recommend using const

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -535,22 +547,21 @@ struct Database {
...
 
-        // mark mutantStatus in mutationStatus-table
-        const s = format!"UPDATE %s SET status=%s WHERE id IN (SELECT st_id FROM %s WHERE id = %s)"(
-                mutationStatusTable, to_status.to!long, mutationTable, id.to!long);
-        auto stmt = db.prepare(s);
-        stmt.execute;
+        if (p.isNull)
+            return;
+
+        auto sl = getSourceLocation(id);
```
joakim-brannstrom: can this fail?

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -535,22 +547,21 @@ struct Database {
...
+        auto r = Rationale(rationale);
```
joakim-brannstrom: `r` and `t` might as well be directly in at line 564 because they can never fail and are only used once.

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -535,22 +547,21 @@ struct Database {
...
+        auto t = Clock.currTime.toUTC;
 
-        // add mutant to separate table for marked mutants
-        db.run(buildSchema!MarkedMutant); // TODO: can probably be done somewhere else
-        db.run(insert!MarkedMutant.insert, MarkedMutant(MutationId(id),
-                Path(getPath(MutationId(id))), from_status, to_status,
-                Rationale(Clock.currTime.toUTC.toSqliteDateTime, rationale)));
+        // update mutationStatus and add mutant to separate table for marked ones.
+        updateMutationStatus(st_id, to_status);
```
joakim-brannstrom: I recommend moving this logic to the caller. Let it decide what to do.
It reduces the encoded business logic in the database and allow a higher degree of flexibility of the caller.
This method might, for example, be used in currently unknown ways where we do not want to changes the status :)

plugin/mutate/source/dextool/plugin/mutate/frontend/argparser.d
```diff
@@ -64,7 +64,7 @@ struct ArgParser {
         ToolMode toolMode;
 
         Mutation.Status to_status;
-        ulong mutant_id;
+        long mutant_id;
```
joakim-brannstrom: this can be removed because it is unused?

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
peatersson: Hmm, I am not sure I completely understand. I pushed more commits, so I just commented it out for now. But do we mean removing it?

peatersson: Ah, you mean move it to the markMutant-function in ../admin.d ? (as an example)

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -83,6 +83,14 @@ struct Database {
         return res.oneValue!long != 0;
     }
 
+    bool hasLostMarkings(Path path) @trusted {
+        auto s = format!"SELECT COUNT(*) FROM %s WHERE path=:path LIMIT 1"(markedMutantTable);
```
joakim-brannstrom: *nagging* change `auto` to `immutable`.

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -84,7 +84,7 @@ struct Database {
     }
 
     bool hasLostMarkings(Path path) @trusted {
-        auto s = format!"SELECT COUNT(*) FROM %s WHERE path=:path LIMIT 1"(markedMutantTable);
+        immutable s = format!"SELECT COUNT(*) FROM %s WHERE path=:path LIMIT 1"(markedMutantTable);
```
joakim-brannstrom: :100:

joakim-brannstrom: I do not understand how this query finds those that have lost a marking. Could you enlighten me?

plugin/mutate/test/test_admin.d
```diff
@@ -0,0 +1,83 @@
...
+    // arrange
+    mixin(EnvSetup(globalTestdir));
+    makeDextoolAnalyze(testEnv).addInputArg(testData ~ "fibonacci.cpp").run;
+
+    // act
+    auto r = makeDextoolAdmin(testEnv)
+        .addArg(["--operation", "markMutant"])
+        .addArg(["--id", "12"])
+        .addArg(["--to-status", "killed"])
+        .addArg(["--rationale", `"A good rationale"`])
```
joakim-brannstrom: hehe :+1:

plugin/mutate/test/test_admin.d
```diff
@@ -0,0 +1,83 @@
...
+        .run;
+
+    // assert
+    testAnyOrder!SubStr(["Failure when marking mutant: 12"]).shouldNotBeIn(r.stderr);
```
joakim-brannstrom: this is very "problematic" tests. there should be at least one that is a positiv test, in this case "shouldBeIn".
By experience a test that only check for the negative case is easy to get "out of sync" with the implementation in some way or another because as long as something do "not" happen it passes.

plugin/mutate/test/test_admin.d
```diff
@@ -0,0 +1,83 @@
...
+        .run;
+    auto secondRes = makeDextoolAdmin(testEnv)
+        .addArg(["--operation", "markMutant"])
+        .addArg(["--id", "3"])
+        .addArg(["--to-status", "unknown"])
+        .addArg(["--rationale", `"Backend was wrong, mutant is legit..."`])
+        .run;
+
+    // assert
+    testAnyOrder!SubStr(["Failure when marking mutant: 3"]).shouldNotBeIn(secondRes.stderr);
```
joakim-brannstrom: same comment here as the previous one regarding negative testing

plugin/mutate/test/utility.d
```diff
@@ -24,6 +24,15 @@ auto makeDextoolAnalyze(const ref TestEnv env) {
     // dfmt on
 }
 
+auto makeDextoolAdmin(const ref TestEnv env) {
+    // dfmt off
+    return dextool_test.makeDextool(env)
+        .setWorkdir(null)
```
joakim-brannstrom: :+1:

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
peatersson: No, that is completely true. I started out making the query more complex, and then slimmed it down to current implementation. Did not change the name that is.

I guess you could say that it WAS how i used it. I will correct it tomorrow.

plugin/mutate/source/dextool/plugin/mutate/backend/database/type.d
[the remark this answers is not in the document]
peatersson: I am currently implementing this, and I have the following:
`
| File | Line | Column | Status | Rationale |
`

Is this really enough for the user, or is something missing?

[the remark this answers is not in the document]
joakim-brannstrom: Yes, mutant kind is probably a good idea. file+line+column+kind makes it globally unique

plugin/mutate/source/dextool/plugin/mutate/backend/analyze/package.d
```diff
@@ -314,13 +314,11 @@ void printLostMarkings(MarkedMutant[] lostMutants) {
...
+    Table!6 tbl = Table!6(["ID", "File", "Line", "Column", "Status", "Rationale"]);
     foreach(m; lostMutants) {
-        typeof(tbl).Row r = [m.path, to!string(m.line), to!string(m.column), statusToString(m.to_status), m.rationale];
+        typeof(tbl).Row r = [to!string(m.id), m.path, to!string(m.line), to!string(m.column), statusToString(m.to_status), m.rationale];
         tbl.put(r);
     }
-
-    writeln;
-    logger.warning("Marked mutants that became obsolete:");
+    logger.warning("Marked mutants was lost");
```
joakim-brannstrom: this should only be printed if there are any lost mutants?

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -83,6 +83,20 @@ struct Database {
         return res.oneValue!long != 0;
     }
 
+    bool lostMarking(long status_id) @trusted {
```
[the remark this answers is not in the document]
peatersson: It is used on row 112 in standalone.d, to filter out the lost markings.

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -433,6 +435,8 @@ struct MarkedMutantTbl {
     string rationale;
 
     ulong kind;
+
+    Language lang;
```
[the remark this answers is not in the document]
peatersson: It was purely added so that creation of MutationTextResult in report/plain.d (row 73) was more reasonable. I agree that the information is not needed for the purpose of displaying the table etc, but instead of doing it the "long way", it was a "nice-to-have". Is it still reasonable?

plugin/mutate/test/test_admin.d
```diff
@@ -168,4 +168,28 @@ unittest {
...
+        .addArg(["--to-status", to!string(toStatus)])
+        .addArg(["--rationale", rationale])
+        .run;
+    auto r = makeDextoolAnalyze(testEnv).addInputArg(testData ~ "abs.cpp").run;
+
+    // assert
+    testAnyOrder!SubStr([
+        "| ID | File                                              | Line | Column | Status           | Rationale |",
+        "|----|---------------------------------------------------|------|--------|------------------|-----------|",
+        "| 3  | build/plugin/mutate/plugin_testdata/fibonacci.cpp | 8    | 10     | killedByCompiler | \"Lost\"    |",
```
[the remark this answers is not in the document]
peatersson: I did at first, but the output in the test, i.e what is captured in r.stdout, had the escaped quotes. Just remade the check so that it worked accordingly.

plugin/mutate/test/test_report.d
```diff
@@ -306,6 +307,35 @@ unittest {
...
+        .addArg(["--section",   "marked_mutants"])
+        .addArg(["--style",     "plain"])
+        .run;
+
+    // assert
+    testConsecutiveSparseOrder!SubStr([
+        "| File                                              | Line | Column | Mutation | Status           | Rationale |",
+        "|---------------------------------------------------|------|--------|----------|------------------|-----------|",
+        "| build/plugin/mutate/plugin_testdata/fibonacci.cpp | 8    | 10     | !=       | killedByCompiler | \"Useless\" |",
+    ]).shouldBeIn(r.stdout);
```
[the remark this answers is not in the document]
peatersson: Haha, always fun.

plugin/mutate/source/dextool/plugin/mutate/backend/analyze/package.d
peatersson: Yes, the check on row 311 should handle that.

plugin/mutate/source/dextool/plugin/mutate/backend/analyze/package.d
```diff
@@ -300,3 +304,21 @@ bool isPathInsideAnyRoot(AbsolutePath[] roots, AbsolutePath f) @safe {
...
+
+    import std.stdio: writeln;
+    import std.conv : to;
+
+    Table!6 tbl = Table!6(["ID", "File", "Line", "Column", "Status", "Rationale"]);
+    foreach(m; lostMutants) {
+        typeof(tbl).Row r = [to!string(m.id), m.path, to!string(m.line), to!string(m.column), statusToString(m.to_status), m.rationale];
+        tbl.put(r);
+    }
+    logger.warning("Marked mutants was lost");
```
joakim-brannstrom: True, I missed it in the small diff

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -411,6 +412,33 @@ struct MutantTimeoutCtxTbl {
...
+
+    ulong to_status;
+
+    SysTime time;
+
+    string rationale;
+
+    ulong kind;
+
+    Language lang;
```
joakim-brannstrom: Hmm actually, I think you need to save the mutant here, the text. Because when it is lost we can't recreate it. The source file has changed

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -83,6 +83,35 @@ struct Database {
         return res.oneValue!long != 0;
     }
 
+    bool isInMutationStatusTable(long status_id) {
```
joakim-brannstrom: This must take as argument a mutationstatusid

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -411,6 +440,17 @@ struct Database {
             rval = MutationStatusId(res.peek!long(0));
         return rval;
     }
+    // Returns: the status of the mutant
+    Nullable!Mutation.Status getMutationStatus(const MutationId id) @trusted {
```
joakim-brannstrom: After thinking about this function and the two other four loc and kind I think they can be removed because all that info is already accessible via getMutation in one call

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
peatersson: Oh, that is absolutely true. I will fix it later today.

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
peatersson: getMutation gives a MutationEntry, which does not contain status or kind. Therefore I created the functions for it, so that it would be easier to access specific items.

joakim-brannstrom:
Do you feel that it is ready to merge?

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -77,6 +77,35 @@ struct Database {
         return res.oneValue!long != 0;
     }
 
+    bool isInMutationStatusTable(MutationStatusId status_id) {
```
joakim-brannstrom: *Note*: avoid leaking implementation details via the API.
*Note*: YAGNI.
I think you should merge isInMutationStatusTable and lostMarking. isInMutationStatusTable is how it is implemented in the database, lostMarking is the external functionality.
Or if you feel that it has wider use in the future I would recommend changing the name to what is externally expected such as "bool exists(MutationStatusId id)".

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -77,6 +77,35 @@ struct Database {
...
+                markedMutantTable, mutationTable);
+        auto stmt = db.prepare(s);
+        stmt.bind(":id", cast(long) id);
+        auto res = stmt.execute;
+        return res.oneValue!long != 0;
+    }
+
+    MarkedMutant[] getLostMarkings() @trusted {
+        import std.algorithm : filter;
+        import std.array : array;
```
joakim-brannstrom: unnecessary import (array)

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -411,6 +412,31 @@ struct MutantTimeoutCtxTbl {
...
 
+import dextool.plugin.mutate.backend.database.type : Rationale;
+
+@TableName(markedMutantTable)
+@TablePrimaryKey("st_id")
+struct MarkedMutantTbl {
+    @ColumnName("st_id")
+    long mutationStatusId;
+
+    long id;
```
joakim-brannstrom: I recommend changing this to mutationId and `@ColumnName("mut_id")` or something. I didn't understand "what" it was an ID so I had to read the source code in standalone.d/markMutant.

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -411,6 +412,31 @@ struct MutantTimeoutCtxTbl {
...
+
+    string path;
+
+    ulong to_status;
+
+    SysTime time;
+
+    string rationale;
+
+    string text;
```
joakim-brannstrom: I recommend changing the variable name to reflect what it is.
Is it the original text? Or the mutated.
I actually do not remember from our conversation so I would have to read how it is used.

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
peatersson: Very true. I could see a use in the future where it is used, so I'll go with the namechange (plus merge that is)

peatersson:
Now I feel ready to merge. Feedback was good as usual, so I updated accordingly. We did agree to add html-report for marked_mutants later, right?

joakim-brannstrom:
Nice man!

Correct regarding the html report.

joakim-brannstrom:
good job @peatersson

PR #959, opened 2019-11-17
mutate: split analyzers and utility functions

joakim-brannstrom:


joakim-brannstrom:
@peatersson , the split should make it a bit clearer what is what.

PR #963, opened 2019-11-17
mutate: add statistics to json report

joakim-brannstrom:


joakim-brannstrom:
this should make it easier to gather statistics from all applications

Issue #902, opened 2019-09-08
Improve the speed when using dextool on pull requests


joakim-brannstrom:
A user wants fast feedback on a pull requests. The user is most probably interested in those mutants that are affected by the changed lines in the pull request.

joakim-brannstrom:
After testing live I deem this closed. Further improvements are incremental on top of this.

PR #966, opened 2019-12-08
mutate: stop mutation testing after a specific number of alive mutants

joakim-brannstrom:
are found in diff mode.


joakim-brannstrom:
Another part for #902

PR #984, opened 2020-01-02
mutate: fix bug when sandbox and timeout combination result in the children surviving

joakim-brannstrom:


peatersson:
First pr/commit of 2020, nice! I was aiming for that tomorrow. :)

PR #990, opened 2020-01-05
mutate: parallel analyze

joakim-brannstrom:


joakim-brannstrom:
It works. Now some cleanup and allow the user to limit how many threads are used.

Issue #942, opened 2019-10-28
Pull Requests


joakim-brannstrom:
**Problem**: How to integrate mutation testing in a pull request workflow?
What is features need to be implemented and integrated together to get a smooth workflow when a user wants to have mutation testing feedback on a pull request.

The users wants to be able to:
 [x] fast feedback (10-20 minutes)
 [x] only report live mutants in the changed code in the pull request
 [x] as few mutants as possible, something like 1-3 mutants.
 [1] *high* impact mutants
 [x] killing a live mutant in the diff and then **not** get new mutants reported for that pull request. the pull request goes *green*.
 [2] mark a live mutant as don't-care (nomut) and then in turn mark the pull request as *green*.

1) not important for now. We will observe the users and see if it is a needed feature in the future.
2) Deemed not necessary for now because of how we have chosen to work with mutation testing

joakim-brannstrom:
After talking with the users today we kind a arrive on maybe a doable approach by instead of focusing on "alive" mutants but rather a mutation score for the pull request.
The user can set a "target" mutation score for the pull request which mean that it either pass or fail.

joakim-brannstrom:
It is starting to be integrated in CI workflow now. It will be interesting to hear what feedback we get.

joakim-brannstrom:
Close because it is good enough to be used as is

PR #998, opened 2020-01-12
mutate: fix performance regression when reporting statistics

joakim-brannstrom:
The improvement for a small example is from 12s to 0,3s.
The call to db.getTestCases(mutationId) was the culprit. It should only be done when it is needed.


joakim-brannstrom:
Important for #942 because it is a "significant" amount of time from the 10 minutes we have.

joakim-brannstrom:
I found out what was the problem by looking at a [flame graph](https://github.com/brendangregg/FlameGraph)

PR #999, opened 2020-01-12
mutate: convert views to tables

joakim-brannstrom:
Slight performance improvement and makes it possible to edit the tables.


joakim-brannstrom:
@peatersson this should make it possible for the mark mutant functionality to be able to also set the NOMUT state for a mutant.

PR #1011, opened 2020-01-20
mutate: fix resource leak

joakim-brannstrom:


joakim-brannstrom:
It was a bad idea to use classes that are handled by the GC for file resources. Converted to structs and RAII.

PR #1012, opened 2020-01-20
mutate: improve analyze speed

joakim-brannstrom:
earlier.


joakim-brannstrom:
There are still significant improvements to be made by tuning sqlite and how dextool uses it.

joakim-brannstrom:
@peatersson , the last commit fixes the bug in the marking of mutants.
the "mutation status id" is reused by sqlite whenever it feels like it. The only "assured to be unique" identifier that can be used is the checksum. I've changed it to using that instead.

PR #1044, opened 2020-03-21
mutate: improve db perf

joakim-brannstrom:


joakim-brannstrom:
This sped up the prune by 25%. Not good enough but something.....

PR #1047, opened 2020-03-22
mutate: speedup prune

joakim-brannstrom:


joakim-brannstrom:
Screenshot_20200322_133235

The progress indication to the user

PR #1094, opened 2020-04-13
mutate: fix sandbox bug

joakim-brannstrom:


joakim-brannstrom:
Generic update from distssh that for dextool fix a bug wherein the sandbox for test cases sometimes could let processes escape.

PR #1135, opened 2020-07-09
mutate: add a undesired mutant filter

joakim-brannstrom:


joakim-brannstrom:
A stepping stone to solve #917

joakim-brannstrom:
This reduces the total number of generated mutants by 10% for the recommended mutation operators.

pokgak:
Is there any docs somewhere on how to use this feature?

joakim-brannstrom:
> Is there any docs somewhere on how to use this feature?

It is automatically used as it is. The remark about about #917 is that with this pull request then it would be relatively simple to add the feature wherein a user would be able to specify <insert magic syntax> a way that informs dextool mutate that the mutant is undesired.

I haven't prioritized it beyond this pull request but if you need it and we can work out a reasonable syntax it should be a matter of hours at the most to implement. I think it would be hard to come up with a configuration syntax for this so I have been leaning towards something like the external analyzer that you used. Such that dextool mutate calls the user-configured analyzer and asks if the mutant is undesired/desired. Dextool would call the external like this:

```sh
external_mutant_filter.py --file foo.cpp --offset-start 42 --offset-end 84 --mutant "true"
```

that would provide maximum flexibility for the user. Then when we find more use cases, the most common "external analyzers" etc it should be possible to integrate them in dextool mutate as built-ins.

pokgak:
I'm using the default operators for now so the mutants generated is still reasonable and I don't need the filter anymore but I can imagine having the ability to write the filter would be useful.

PR #1150, opened 2020-08-05
mutate: fix html colors

christoffer-nylen:
1) This fixes the problem that light pink (alive) vs mistyrose (unknown) can be easy to mix by mistake, by introducing a light yellow color for unknown.

2) It changes time-out mutants to use the same color brightness as killed mutants (light green), to reduce confusion.

image


joakim-brannstrom:
Looking good

Issue #1165, opened 2020-08-17
mutate: show selected mutant 


pokgak:
When I ran `dextool mutate test`, I don't know which kind of mutant is tested. There is also no documentation on what the default mutant is when the user does not pass the `--mutant` option to the command. The same issue are also present for `report` and (maybe) `admin`, which caused me some confusion at first thinking why the number is so low.

So maybe something like the following:
```
info: Using /home/pokgak/git/RIOT-mutate/mutate/dextool_mutate.sqlite3
info: Summary
info: mutants: lcr, ror        <-------------- something like this
Time spent:  2 hours, 16 minutes, 54 secs, and 442 ms
Time spent on mutants killed by compiler: 2 secs and 675 ms

Score:       0.547
Total:       225
Alive:       102
Killed:      123
Timeout:     0
Killed by compiler: 5
```

joakim-brannstrom:
Maybe #1166 will help?

pokgak:
Yup, that does it. Never knew the default uses that many operators. I thought it is only the lcr. Thanks for your work :)

joakim-brannstrom:
> Yup, that does it. Never knew the default uses that many operators. I thought it is only the lcr. Thanks for your work :)

Thank you.

Yes, the default is based on our experience from using it on a couple of hundred applications. The priority that we, subjectively so far, have come up with based on the relevance of surviving mutants is
lcr>dcr>sdl>aor>rorp

uoi has been added lately. Stilll evaluating its relevance. But the restricted implementation of only deleting occurences of "!" (negation) seems to be pretty spot on.

PR #1166, opened 2020-08-17
mutate: print what mutation operators are used

joakim-brannstrom:
This isn't all entry points but the most common.


joakim-brannstrom:
For #1165

PR #1185, opened 2020-09-26
Fix for latest dmd

joakim-brannstrom:


peatersson:
Seems legit, works! :)

PR #1202, opened 2020-11-17
mutate: fix sorting in html report

joakim-brannstrom-work1:


joakim-brannstrom-work1:
this works in firefox but not Konqueror (KHTML). but it is better with a partial solution that make it a bit better than before

PR #1234, opened 2020-12-14
mutate: add more sections to json report

joakim-brannstrom-work1:


joakim-brannstrom-work1:
mylib is failing on CI, will fix later

PR #1257, opened 2020-12-26
mutate: separate compile test times

joakim-brannstrom:


joakim-brannstrom:
That should do it. Improved time accounting. Now even failing to compile a schemata is recorded.

PR #1261, opened 2020-12-29
mutate: add function coverage

joakim-brannstrom:


joakim-brannstrom:
Database schema, source code analysis and saving the result in the database is done and working. Instrumentation and gathering the data tomorrow.

joakim-brannstrom:
The approach, as I have done it, do allow it to be extended to gather information about all visited blocks if needed in the future. But I am keeping it at the function level for now. Simple is best to begin with :)

plugin/mutate/source/dextool/plugin/mutate/backend/analyze/pass_coverage.d
```diff
@@ -0,0 +1,114 @@
...
+    private void put(const AbsolutePath p, const Interval i) {
+        if (auto v = p in points) {
+            *v ~= i;
+        }
+    }
+}
+
+private:
+
+class CoverageVisitor : DepthFirstVisitor {
```
peatersson, line 85: Pretty much spot on what we discussed earlier. Seems very reasonable.

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -548,6 +551,41 @@ struct MutationScoreHistoryTable {
...
+@TableConstraint("file_offset UNIQUE (file_id, begin, end)")
+struct CoverageCodeRegionTable {
+    long id;
+
+    @ColumnName("file_id")
+    long fileId;
+
+    /// the region in the files, in bytes.
+    uint begin;
+    uint end;
```
peatersson, line 568: Hmm, so this will be all the functions for one file, but saved separately in the db.

Could it be a problem with a lot of db entries in the end?

peatersson: Nevermind, I was thinking of a solution where each row in the table is growing with every function, but that does not seem more reasonable at all.

joakim-brannstrom: > Hmm, so this will be all the functions for one file, but saved separately in the db.
> 
> Could it be a problem with a lot of db entries in the end?

that is correct but the data should be pretty minimal as you see

joakim-brannstrom:
generating the instrumented source code now. left to implement the shared memory and then reading the result to save it in the database.

PR #1309, opened 2021-01-29
mutate: update sanity checks

joakim-brannstrom:


joakim-brannstrom:
@christoffer-nylen this contains the write permission check and a check for the build/test environment

christoffer-nylen:
Najs!! 😁

PR #1317, opened 2021-02-03
mutate: save testcmd as tc

joakim-brannstrom:


joakim-brannstrom:
@christoffer-nylen now you can experiment with adding test_cmds and dextool will automatically re-execute alive + save which test_cmd killed the mutant.

PR #1364, opened 2021-04-24
mutate: reduce pointer pessimism for schematan

joakim-brannstrom:
The overly pessimistic blocking of schematan when any sub node is a
pointer meant that e.g. x->foo() where blocked which lovered the speed.


joakim-brannstrom:
superseded by #1459

PR #1388, opened 2021-05-16
mutate: fix lockup if a test_cmd contains test cases with the same name

joakim-brannstrom:


joakim-brannstrom:
assumptions meet reality.... <boom>

PR #1398, opened 2021-05-24
mutate: fix checksum check

joakim-brannstrom-work1:


joakim-brannstrom-work1:
This wasn't a problem but I prefer this. A bit of extra checks and more debugging to make it possible to follow what happens during the test phase.

PR #1478, opened 2021-08-29
mutate: split fragments based on probability

joakim-brannstrom:


joakim-brannstrom:
still evaluating if this is a good change or not. but will merge because it contains some critical fixes too and is a nice refactoring.

Issue #1537, opened 2022-01-29
Facing trouble in installation


meng004:
Installation environment: Ubuntu 20.04
According to [Ubuntu.md](https://github.com/joakim-brannstrom/dextool/blob/master/doc/install/UBUNTU.md) and [Readme.md](https://github.com/joakim-brannstrom/dextool/blob/master/README.md#build-and-install), several troubles could not be addressed by common build errors.
1 using instrument "dub run dextool", error message "/home/vagrant/.dub/packages/dextool-4.0.0/dextool/vendor/undead/src/undead/xml.d(594,49): Error: scope variable `o` assigned to non-scope parameter `o` calling undead.xml.toType!(const(Document)).toType" and "[CMakeFiles/Makefile2:797: vendor/CMakeFiles/dextool_undead.dir/all] Error 2" are shown while " Generating dextool_undead.o"
![image](https://user-images.githubusercontent.com/4920208/151644413-eb0e155e-56b3-427e-88fe-02735b776367.png)

2 executing "make install" from source code, error message "/home/vagrant/dextool/vendor/mylib/source/my/actor/actor.d(69,34): Error: copy constructor my.actor.mailbox.Reply.this(return ref Reply a) is not callable using argument types (ulong, VariantN!32LU)
/home/vagrant/dextool/vendor/mylib/source/my/actor/actor.d(69,34):        cannot pass argument this.data.get().replyId of type ulong to parameter return ref Reply a" and " [CMakeFiles/Makefile2:933: vendor/CMakeFiles/dextool_mylib.dir/all] Error 2" are illustrated while "Generating dextool_mylib.o"
![image](https://user-images.githubusercontent.com/4920208/151644383-0487574d-082f-49ae-b2ee-c65ae3e077b8.png)

Thanks for your kind help.

meng004:
I have solved this issue by CMake.
The key step is that install the D compiler following [official instruments](https://dlang.org/install.html#install).

All of the scripts are listed as follows. I hope it can provide enlightening information to other developers.

` sudo apt install build-essential cmake llvm-13 llvm-13-dev clang-13 libclang-13-dev libsqlite3-dev`

`mkdir -p ~/dlang
wget https://dlang.org/install.sh -O ~/dlang/install.sh
sudo chmod +777 ~/dlang/install.sh
~/dlang/install.sh install dmd-2.097.2
~/dlang/install.sh install ldc-1.28.0 
~/dlang/install.sh install dub 

[...]

PR #1538, opened 2022-02-02
readme: update instructions for ubuntu

joakim-brannstrom:


joakim-brannstrom:
#1537 add instructions from meng004

PR #1563, opened 2022-06-16
Trend

virrenn:
Now displays the Trend graph as individual files, where the y axis is the mutation score and the x axis is the time when the tests were run. Also added an "Average" line for all the files. If you press the name of the file at the top it's line is removed from the graph. 

**Imporvements that should be fixed**:
**report/json.d:** Should loop over values (Not working as intended currently)
**report/plain.d:** Should loop over values (Not working as intended currently) 
**backend/test_mutant/package.d:** _string[] files = ["....]_ is hardcoded, should get the file paths from the database via **db.getFilesStrings()**, but I am having difficulties with "@safe nothrow"

### DIFF
--- a/plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
+++ b/plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
@@ -653,8 +653,8 @@ struct MutationScoreHistoryTable {
 
     double score;
 
-    @ColumnName("file_id")
-    long fileId;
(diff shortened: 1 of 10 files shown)


virrenn:
Har fixat så att "mutate report --style plain --section trend" visar all data nu, dock får du gärna säga om du vill att den ska visas finare, typ att man sorterar enligt filnamn och tid, och endast visar filnamn en gång, blir lite lättare att läsa då

PR #1564, opened 2022-06-16
Backarrow fix

OliverGreen27:
Now the backarrow works to actually go to previous-page. removed anchors when clicking on mutants etc


joakim-brannstrom-work1:
Lets merge! From our discussion yesterday it shouldn't break anything and if it happens to do that we have `git revert`.

PR #1565, opened 2022-06-16
mutate: Only show mutants with unique statuses on the same line

virrenn:
If there are more than one mutant with the same status on the same line in the HTML report, only one of them is displayed


plugin/mutate/source/dextool/plugin/mutate/backend/database/package.d
```diff
@@ -279,6 +279,7 @@ struct Database {
             t0.mp_id = t1.id AND
             t1.file_id = t2.id AND
             t2.path = :path
+            GROUP BY line, status
```
joakim-brannstrom-work1, line 282: How will this work for mutants that are of the same kind, are on the same line and have the same status but mutate the code in unique ways?
Wouldn't this mean that only one of those are shown?

PR #1569, opened 2022-06-20
mutate: trend per file

virrenn:
Added a column in the Files section that shows the difference in the mutation score for a file compared to seven days ago.

A problem with this would be when people first start using it there are no scores from more than seven days ago, therefore it sets that score to zero. This means that for the first week the "score change" would always show: (current score) - 0, which is untrue for projects that have used this for more than a week. A possible solution would be not showing the "score change" if there are no scores from more than seven days ago.


virrenn:
A problem with this would be when people first start using it there are no scores from more than seven days ago, therefore it sets that score to zero. This means that for the first week the "score change" would always show: (current score) - 0, which is untrue. A possible solution would be not showing the "score change" if there are no scores from more than seven days ago.

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -479,14 +531,15 @@ void toIndex(FileIndex[] files, Element root, string htmlFileDir) @trusted {
             "form-control").setAttribute("placeholder", "Search...");
 
     auto tbl = tmplSortableTable(root, [
-            "Path", "Score", "Alive", "NoMut", "Total", "Time (min)"
+            "Path", "Score", "Change (7 days)", "Alive", "NoMut", "Total", "Time (min)"
```
joakim-brannstrom-work1, line 534: Maybe add a [?] with extra info and keep the column name "short"

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -511,8 +564,26 @@ void toIndex(FileIndex[] files, Element root, string htmlFileDir) @trusted {
                     return "background-color: lightgreen";
                 return null;
             }();
-
             r.addChild("td", format!"%.3s"(score)).style = style;
+
+            float scoreChange;
```
joakim-brannstrom-work1, line 569: use double

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -511,8 +564,26 @@ void toIndex(FileIndex[] files, Element root, string htmlFileDir) @trusted {
...
+            if(f.display in scoreDifference){
+              scoreChange = scoreDifference[f.display];
+            }else{
+              scoreChange = 0;
+            }
+
+            const scoreChangeStyle = () {
+                if (scoreChange < 0)
+                    return "background-color: salmon";
+                if (scoreChange == 0)
```
joakim-brannstrom-work1, line 579: need to use some kind of threshold because it will never be `0`.
see impl for prediction of trend.

joakim-brannstrom-work1: it may be possible, should be possible, to reuse that impl by separating it to `utility.d`.
the colors should also be possible to reuse. one color schema to rule them all

plugin/mutate/source/dextool/plugin/mutate/backend/test_mutant/package.d
```diff
@@ -1113,16 +1113,29 @@ nothrow:
...
+            db.putMutationScore(MutationScore(time, typeof(MutationScore.score)(score.score)));
             db.trimMutationScore(10000);
             t.commit;
         });
+
+        foreach(fileScore; fileScores){
+            spinSql!(() @trusted {
+                auto t = db.transaction;
+                db.putMutationFileScore(MutationScore(time, typeof(MutationScore.score)(fileScore.score), fileScore.filePath));
+                db.trimMutationScore(10000);
```
joakim-brannstrom-work1, line 1135: need a separate db query. this one will not affect the new table

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -892,6 +907,7 @@ void upgradeV0(ref Miniorm db) {
             SchemataMutantTable,
             SchemataUsedTable, SchemaMutantKindQTable, SchemaSizeQTable,
             MutantWorklistTbl, RuntimeHistoryTable, MutationScoreHistoryTable,
+            MutationFileScoreHistoryTable,
```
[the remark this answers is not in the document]
virrenn: Do I need to add any execution of the upgrade? Or is that supposed to be done by the "user"?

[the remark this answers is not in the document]
joakim-brannstrom-work1: the execute is automatically detected. at compile time `makeUpgradeTable` inspect the module for functions named `upgradeV*`

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
virrenn: > need to use some kind of threshold because it will never be `0`. see impl for prediction of trend.

Currently it will be zero is there is no change in score in the last seven days, should it not be shown at all if it is zero instead?

plugin/mutate/source/dextool/plugin/mutate/backend/database/type.d
```diff
@@ -322,6 +322,12 @@ struct MutationScore {
     NamedType!(double, Tag!"MutationScore", 0.0, TagStringable) score;
 }
 
+struct FileScore {
+    SysTime timeStamp;
+    NamedType!(double, Tag!"MutationScore", 0.0, TagStringable) score;
```
joakim-brannstrom, line 327: tag should be `MutationFileScore`, then we get a compilation error if they are accidentally mixed with `MutationScore`

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -280,12 +280,50 @@ struct Database {
         return app.data.sort!((a, b) => a.timeStamp < b.timeStamp).array;
     }
 
+    /// Returns: the stored scores in ascending order by their `time`.
+    FileScore[] getMutationFileScoreHistory() @trusted {
```
joakim-brannstrom, line 284: instead of retrieving all it seems like it is only those with a timestamp > 7days that are of interest.
i recommend adding that as a parameter, modify the query and let SQL sort them by date. then most D logic and cleanup are captured by the query

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/tmpl.d
```diff
@@ -76,8 +76,14 @@ Table tmplSortableTable(Element n, string[] header) @trusted {
     auto tr = base.div.parentDocument.createElement("tr");
     foreach (h; header.enumerate) {
         auto th = tr.addChild("th", h.value);
-        DashboardCss.sortableTableCol(th).setAttribute("id",
-                format!"col-%s"(h.index)).appendText(" ").addChild("i").addClass("right");
+        if (h.value == "Change") {
```
joakim-brannstrom-work1, line 79: tmplSortableTable is generic thus it can't have specific knowledge about how columns in other parts of dextool should be formatted. change tmplSortableTable to e.g. taking a callback as a parameter. then you can implement this logic at the callsite.

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -1439,8 +1527,10 @@ auto spawnOverviewActor(OverviewActor.Impl self, FlowControlActor.Address flowCt
         if (ReportSection.treemap in ctx.state.get.sections) {
             addSubPage(() => makeTreeMapPage(ctx.state.get.files), "tree_map", "Treemap");
         }
+        import dextool.plugin.mutate.backend.database.type : FileScore;
 
-        ctx.state.get.files.toIndex(content, HtmlStyle.fileDir);
+        FileScore[] fileScores = ctx.state.get.db.getMutationFileScoreHistory();
```
joakim-brannstrom-work1, line 1532: use auto instead of `FileScore[]`

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -479,15 +534,30 @@ void toIndex(FileIndex[] files, Element root, string htmlFileDir) @trusted {
             "filter_table_on_search('fileFilterInput', 'fileTable')").addClass(
             "form-control").setAttribute("placeholder", "Search...");
 
-    auto tbl = tmplSortableTable(root, [
-            "Path", "Score", "Alive", "NoMut", "Total", "Time (min)"
-            ]);
+    Table tbl;
```
joakim-brannstrom-work1, line 537: line 537 to 548 is initialization of tbl. I recommend using a lambda for the init.

```d
Table tbl = () { if (x) return tbl.init else return tblAwesome; }();
```

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -280,12 +280,63 @@ struct Database {
...
+    void trimFileScore(const long keep, Path file) @trusted {
+        auto stmt = db.prepare(format!"SELECT count(*) FROM %s WHERE file_path=:file"(mutationFileScoreHistoryTable));
+        stmt.get.bind(":file", file.toString);
+        const sz = stmt.get.execute.oneValue!long;
+
+        if (sz < keep)
+            return;
+
+        auto ids = appender!(long[])();
+        stmt = db.prepare(format!"SELECT id FROM %s WHERE file_path=:file ORDER BY time_stamp ASC LIMIT :limit"(
```
joakim-brannstrom-work1, line 325: I think for sqlite to be able to order datetime columns correctly you need to use `datetime(time_stamp)` for it to understand it is "time" and not something else.

PR #1570, opened 2022-06-21
Priority hightlight

OliverGreen27:
clicking on mutant in file report with mouse now shows the smallest by text length mutant. List of mutants in info-box is also sorted by smallest text length

### DIFF
--- a/plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
+++ b/plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
@@ -21,6 +21,7 @@ import std.range : only;
 import std.stdio : File;
 import std.typecons : tuple, Tuple;
 import std.utf : toUTF8, byChar;
+import std.conv;
 
 import arsd.dom : Document, Element, require, Table, RawSource, Link;
 import my.actor;
@@ -32,6 +33,7 @@ import my.set;
 import dextool.plugin.mutate.backend.database : Database, FileRow,
     FileMutantRow, MutationId, MutationStatusId;
 import dextool.plugin.mutate.backend.database.type : CovRegionStatus;
+
 import dextool.plugin.mutate.backend.diff_parser : Diff;
 import dextool.plugin.mutate.backend.interface_ : FilesysIO;
 import dextool.plugin.mutate.backend.report.type : FileReport, FilesReporter;
@@ -327,7 +329,6 @@ struct Spanner {
 @("shall be possible to construct a Spanner in @safe")
 @safe unittest {
     import std.algorithm;
-    import std.conv;
     import std.range;
     import clang.c.Index : CXTokenKind;
 
@@ -433,7 +434,6 @@ struct Span {
 @("shall return a range grouping mutants by the tokens they overlap")
 @safe unittest {
     import std.algorithm;
-    import std.conv;
     import std.range;
     import clang.c.Index : CXTokenKind;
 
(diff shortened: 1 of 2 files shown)


plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -566,21 +566,29 @@ struct MetaSpan {
 
     StatusColor status;
     string onClick;
+    static struct T {
+        ulong length = -1;
+        MutationId id;
+        ubyte status = Mutation.Status.killed;
```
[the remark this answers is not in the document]
OliverGreen27: is it possible to reference the specific enum type from the type.d file then? or do I use enum StatusColor that also recides in MetaSpan and just reorder the values(alive,kiled, etc...) cause they are in the wrong order

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -564,23 +564,38 @@ struct MetaSpan {
...
+        if (muts.length != 0) {
+            clickPrio.length = muts[0].txt.mutation.length;
+            clickPrio.id = muts[0].id;
+            clickPrio.status = muts[0].mut.status;
+        }
         foreach (ref const m; muts) {
             status = pickColor(m, status);
-            if (onClick.length == 0 && m.mut.status == Mutation.Status.alive) {
-                onClick = format(click_fmt2, m.id.get);
+            if (m.mut.status == clickPrio.status
```
joakim-brannstrom, line 588: Need to impl the previous algo:


1. priorities alive mutants over everything else
2. priorities the shortest alive
3. priorities shortest mutant (which mean that if there are no alive mutants on the line then it is only the size)

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -564,23 +564,37 @@ struct MetaSpan {
         noCoverage,
     }
 
+    static struct MutationLength {
+        ulong length;
+        MutationId id;
```
joakim-brannstrom-work1, line 569: should probably be MutationStatusId considering Viktors changes?

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -564,23 +564,37 @@ struct MetaSpan {
...
+        if (muts.length != 0) {
+            clickPrio = MutationLength(muts[0].txt.mutation.length, muts[0].id, muts[0].mut.status);
+        }
         foreach (ref const m; muts) {
             status = pickColor(m, status);
-            if (onClick.length == 0 && m.mut.status == Mutation.Status.alive) {
-                onClick = format(click_fmt2, m.id.get);
+            if (m.mut.status == Mutation.Status.alive && clickPrio.status != Mutation.Status.alive) {
+                clickPrio = MutationLength(m.txt.mutation.length, m.id, m.mut.status);
+            } else if (m.txt.mutation.length < clickPrio.length && m.mut.status == clickPrio.status) {
```
joakim-brannstrom-work1, line 593: questions.
1. the desired behavior is that the smallest alive is shown. i think the impl handle that.
2. if there are no alive then the desired is that the smallest mutant is shown irregardless of status.
- i do not think the impl that that case?

OliverGreen27: ok so just removing the last statement in the elseif statement would fix that right?

OliverGreen27: the algo was wrong, testing helped to build it better. Thank you!

PR #1581, opened 2022-07-01
mutate: csp error info added

axelbostrom:
If javascript isn't loaded a information text will show

### DIFF
--- a/plugin/mutate/data/dashboard.html
+++ b/plugin/mutate/data/dashboard.html
@@ -12,6 +12,16 @@
   </head>
 
   <body>
+    <div id ="csp_error">
+      <h2>Something isn't right...</h2>
+      <p>To be able to read the mutation testing report you have to turn off CSP in Firefox.</p>
+      <ol>
+        <li>Copy and paste the following into the address bar in firefox: about:config.</li>
+        <li>Search for "CSP".</li>
+        <li>Change "security.CSP.enabled" from true to false.</li>
+        <li>Reload this page.</li>
+      </ol>
+    </div>
     <nav class="navbar navbar-inverse navbar-fixed-top">
       <div class="container-fluid">
         <div class="navbar-header">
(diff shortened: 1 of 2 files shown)


plugin/mutate/data/dashboard.html
```diff
@@ -9,6 +9,17 @@
...
 
     <title></title>
+    
+    <div id ="csp_error">
+      <h2>Something isn't right...</h2>
+      <p>To be able to read the mutation testing report you have to turn off CSP in Firefox.</p>
+      <ol>
+        <li>Copy and paste the following into the address bar in firefox: about:config.</li>
+        <li>Search for "CSP".</li>
+        <li>Change "security.OCSP.enabled" from true to false.</li>
```
joakim-brannstrom-work1, line 19: spelling? OCSP -> CSP?

plugin/mutate/data/dashboard.html
```diff
@@ -9,6 +9,17 @@
     <meta name="author" content="Joakim Brännström">
 
     <title></title>
+    
+    <div id ="csp_error">
```
joakim-brannstrom-work1, line 13: the div tag should be inside the body tag?

plugin/mutate/data/dashboard.html
```diff
@@ -12,6 +12,16 @@
   </head>
 
   <body>
+    <div id ="csp_error">
+      <h2>Something isn't right...</h2>
```
joakim-brannstrom, line 16: Maybe add what the problem is.
"Something isn't right... Javascript need to be enabled"

plugin/mutate/data/dashboard.html
```diff
@@ -12,6 +12,16 @@
...
+      <p>To be able to read the mutation testing report you have to turn off CSP in Firefox.</p>
```
joakim-brannstrom, line 17: If you are using Firefox there may be a problem with Content Security Protection (CSP) with the web server that is used to publish the mutation report. To see the report correctly you need to turn off CSP. Follow these steps:

PR #1587, opened 2022-07-04
mutate: trend-per-file (2)

virrenn:
Made a new pull request since I was unable to squash the commits cleanly on the first PR

### DIFF
--- a/plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
+++ b/plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
@@ -280,10 +280,61 @@ struct Database {
         return app.data.sort!((a, b) => a.timeStamp < b.timeStamp).array;
     }
 
+    /// Returns: the stored scores in ascending order by their `time`.
+    FileScore[] getMutationFileScoreHistory() @trusted {
+        auto app = appender!(FileScore[])();
+        auto stmt = db.prepare(format!"SELECT AVG(score), file_path FROM (SELECT * FROM %s WHERE time_stamp > date('now', '-7 day')) GROUP BY file_path"(
+                mutationFileScoreHistoryTable));
+        foreach (a; stmt.get.execute) {
+            app.put(FileScore(SysTime(0),
+                    typeof(FileScore.score)(a.peek!double(0)), Path(a.peek!string(1))));
+        }
+        return app.data;
+    }
+
     /// Add a mutation score to the history table.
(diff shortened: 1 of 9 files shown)


plugin/mutate/source/dextool/plugin/mutate/backend/report/html/tmpl.d
```diff
@@ -79,8 +79,15 @@ Table tmplSortableTable(Element n, string[] header, void delegate(Element e,
         auto th = tr.addChild("th", h.value);
         if (tdCallback)
             tdCallback(th, h.value);
-        DashboardCss.sortableTableCol(th).setAttribute("id",
-                format!"col-%s"(h.index)).appendText(" ").addChild("i").addClass("right");
+
+        if (h.value == "Change") {
```
joakim-brannstrom-work1, line 83: change the function to taking a callback with the header as argument

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -280,12 +280,63 @@ struct Database {
...
+        auto stmt = db.prepare(format!"SELECT count(*) FROM %s WHERE file_path=:file"(
+                mutationFileScoreHistoryTable));
+        stmt.get.bind(":file", file.toString);
+        const sz = stmt.get.execute.oneValue!long;
+
+        if (sz < keep)
+            return;
+
+        auto ids = appender!(long[])();
+        stmt = db.prepare(format!"SELECT id FROM %s WHERE file_path=:file ORDER BY time_stamp ASC LIMIT :limit"(
```
joakim-brannstrom-work1, line 325: curious question. i wonder if it is possible to write an sql query that delete all ID's that this query return? then the loop at 332-337 wouldn't be needed.

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/constants.d
```diff
@@ -56,7 +57,11 @@ struct DashboardCss {
         return root.addClass("table-sortable");
     }
 
-    static Element sortableTableCol(Element root) @trusted {
-        return root.addClass("table-col-sortable");
+    static Element sortableTableCol(Element root, string headerValue = "") @trusted {
+        if (headerValue == "" || headerValue != "Change") {
```
joakim-brannstrom, line 61: the condition for "Change" need to be moved to the caller.

PR #1592, opened 2022-07-04
mutate: Hover mutant

virrenn:
Shows a tooltip above the hovered mutant

### DIFF
--- a/plugin/mutate/data/source.js
+++ b/plugin/mutate/data/source.js
@@ -260,10 +260,18 @@ function make_kind_status_info(locid, curr_mutid) {
         if (!g_filter_kinds.includes(mut.kindGroup) &&
             !g_filter_status.includes(mut.status))
         {
+            var mut_kind = g_mut_kind_map[mut.kind];
+            var mutant_info_text = "";
+            if (mut_kind == 'stmtDel'){
+                mutant_info_text = "Remove statement";
+            } else {
+                mutant_info_text = mut.mutText;
+            }
+
             if (mutids[i]==curr_mutid)
-                html+=`<span class="hover_`+g_mut_st_map[mut.status]+`"><b><u>`+g_mut_kind_map[mut.kind]+`</b></u> <span class="mutant_info_text">`+g_mut_kind_map[mut.kind] +`</span></span>`;
+                html+=`<span class="hover_`+g_mut_st_map[mut.status]+`"><b><u>`+mut_kind+`</b></u> <span class="mutant_info_text">`+ mutant_info_text +`</span></span>`;
             else
-                html+=`<span class="hover_`+g_mut_st_map[mut.status]+`">`+g_mut_kind_map[mut.kind]+` <span class="mutant_info_text">`+g_mut_kind_map[mut.kind] +`</span></span>`;
+                html+=`<span class="hover_`+g_mut_st_map[mut.status]+`">`+mut_kind+` <span class="mutant_info_text">`+mutant_info_text+`</span></span>`;
         }
     }
     return html;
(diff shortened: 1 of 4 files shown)


plugin/mutate/data/source.js
```diff
@@ -260,10 +260,18 @@ function make_kind_status_info(locid, curr_mutid) {
         if (!g_filter_kinds.includes(mut.kindGroup) &&
             !g_filter_status.includes(mut.status))
         {
+            var mut_kind = g_mut_kind_map[mut.kind];
+            var mutant_info_text = "";
+            if (mut_kind == 'stmtDel'){
```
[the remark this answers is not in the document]
joakim-brannstrom: I think it is enough if there are comments for those in enum type `MutationKind`. That is what the users asked for, at a mininum. But then you need to also generate a json dict that translate a `Mutation.Kind`to a `MutationKind`. You can find the helper functions for this in `mutation_type/package.d`.

plugin/mutate/source/dextool/plugin/mutate/backend/mutation_type/package.d
```diff
@@ -64,9 +64,15 @@ MutationKind toUser(Mutation.Kind k) @safe nothrow {
...
 
+//Convert kind to its description
+string toDescription(MutationKind kind) @safe nothrow{
+    return mutationDescription[kind];
+}
+
 private:
 
 immutable MutationKind[Mutation.Kind] fromInteralKindToUserKind;
+immutable string[MutationKind] mutationDescription;
```
joakim-brannstrom, line 75: might as well make it public. then the toDescription function isn't needed.
then when iterating over the keys you can use ".byKeyValue" instead of EnumMembers

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -730,6 +730,18 @@ void generateFile(ref Database db, ref FileCtx ctx) @trusted {
 
     line.addChild("span", "1:").addClass("line_nr");
     auto mut_data = appender!(string[])();
+
+    
+
+
+    mut_data.put("var g_mut_description = {};");
+    mut_data.put(format!"g_mut_description[%s] = %s;"("''", "'Undefined'"));
```
joakim-brannstrom, line 738: might as well hard code the string instead of using `format`

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -730,6 +730,18 @@ void generateFile(ref Database db, ref FileCtx ctx) @trusted {
...
+    foreach(kind; [EnumMembers!MutationKind]){
```
joakim-brannstrom, line 739: instead of the condition in the loop i recommend using `filter`.

```d
foreach (kind; mutationDescription.byKeyValue.filter!(a => a.key != MutationKind.all)) {..}
```

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -730,6 +730,18 @@ void generateFile(ref Database db, ref FileCtx ctx) @trusted {
 
     line.addChild("span", "1:").addClass("line_nr");
     auto mut_data = appender!(string[])();
+
```
joakim-brannstrom, line 733: too many empty newlines

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/package.d
```diff
@@ -730,6 +730,18 @@ void generateFile(ref Database db, ref FileCtx ctx) @trusted {
...
+    
+
+
+    mut_data.put("var g_mut_description = {};");
+    mut_data.put(format!"g_mut_description[%s] = %s;"("''", "'Undefined'"));
+    foreach(kind; [EnumMembers!MutationKind]){
+        if(kind != MutationKind.all){
```
joakim-brannstrom, line 740: run autoformat

PR #1593, opened 2022-07-04
Mutate: treemap removed

OliverGreen27:
Treemap removed everywhere except d3.min.js


joakim-brannstrom:
Also remove tree_map from enum type `ReportSection`

PR #1596, opened 2022-07-04
Mutate: Removal of tree_map

OliverGreen27:


joakim-brannstrom:
`resource.d` contains `treemap.js` that need to be removed too.

PR #1598, opened 2022-07-05
mutate: style fix js files

axelbostrom:


joakim-brannstrom:
do not run on treemap.js. it is to be removed

PR #1611, opened 2022-07-08
mutate: extend NOMUT support

joakim-brannstrom:


joakim-brannstrom:
I want to include this feature in the next release so I did the last fixes in #1611 . I've set you as author so no credit lost (I hope).

PR #1614, opened 2022-08-08
mutate: Only one mutation-score per day

virrenn:
Need to run autoformat, but the format changes a lot so it should maybe be run in a different PR?

### DIFF
--- a/plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
+++ b/plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
@@ -641,6 +641,7 @@ struct RuntimeHistoryTable {
 }
 
 @TableName(mutationScoreHistoryTable)
+@TableConstraint("time UNIQUE (time)")
 struct MutationScoreHistoryTable {
     long id;
 
@@ -652,6 +653,7 @@ struct MutationScoreHistoryTable {
 }
 
 @TableName(mutationFileScoreHistoryTable)
+@TableConstraint("date UNIQUE (time_stamp, file_path)")
 struct MutationFileScoreHistoryTable {
     long id;
 
@@ -893,6 +895,17 @@ void upgrade(ref Miniorm db, UpgradeTable tbl) {
  */
 void upgradeV0(ref Miniorm db) {
     auto tbl = makeUpgradeTable;
+
+    @TableName(mutationScoreHistoryTable)
+    struct MutationScoreHistoryTable {
+        long id;
+
+        /// when the measurement was taken.
+        @ColumnName("time")
+        SysTime timeStamp;
+
+        double score;
+    }
 
     db.run(buildSchema!(VersionTbl, RawSrcMetadata, FilesTbl,
             MutationPointTbl, MutationTbl, TestCaseKilledTbl, AllTestCaseTbl,
@@ -1996,6 +2009,20 @@ void upgradeV51(ref Miniorm db) {
 
 // 2022-06-21
 void upgradeV52(ref Miniorm db) {
+    @TableName(mutationFileScoreHistoryTable)
+    struct MutationFileScoreHistoryTable {
+        long id;
+
+        /// when the measurement was taken.
+        @ColumnName("time_stamp")
+        SysTime timeStamp;
+
+        double score;
+
+        @ColumnName("file_path")
+        string filePath;
+    }
+
     db.run(buildSchema!MutationFileScoreHistoryTable);
 }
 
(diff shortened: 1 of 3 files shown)


joakim-brannstrom:
linking error in the test cases

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -641,23 +641,25 @@ struct RuntimeHistoryTable {
...
 
 @TableName(mutationScoreHistoryTable)
+@TableConstraint("time UNIQUE (time)")
 struct MutationScoreHistoryTable {
     long id;
 
     /// when the measurement was taken.
     @ColumnName("time")
-    SysTime timeStamp;
+    string timeStamp;
```
joakim-brannstrom, line 650: It is preferable to keep it as a SysTime because, it is time. type safety :)
Have you considered that instead of changing it to a string changing how the systime is saved? such that it is still a SysTime but hours:minues:seconds:usecs are set to zero, always, when saving to the database. That togethered with a unique constraint may be good enough?

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -894,6 +896,17 @@ void upgrade(ref Miniorm db, UpgradeTable tbl) {
 void upgradeV0(ref Miniorm db) {
     auto tbl = makeUpgradeTable;
 
+    @TableName(mutationScoreHistoryTable)
```
joakim-brannstrom-work1, line 899: this should be in upgradeV24

plugin/mutate/source/dextool/plugin/mutate/backend/database/schema.d
```diff
@@ -2032,6 +2059,22 @@ void upgradeV53(ref Miniorm db) {
     replaceTbl(db, newMarkedMutantTbl, markedMutantTable);
 }
 
+// 2022-08-17
+void upgradeV54(ref Miniorm db) {
+    static immutable newFileScoreTable = "new_" ~ mutationFileScoreHistoryTable;
+    db.run(buildSchema!MutationFileScoreHistoryTable("new_"));
+    db.run("INSERT INTO " ~ newFileScoreTable
+            ~ " (id,time_stamp,score,file_path) SELECT id,date(time_stamp) AS time,score,file_path FROM "
+            ~ mutationFileScoreHistoryTable ~ " GROUP BY date(time), file_path");
```
joakim-brannstrom-work1, line 2068: do this query work?
e.g. the select part have different number of columns.

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -292,14 +292,43 @@ struct Database {
 
     /// Add a mutation score to the history table.
     void putMutationScore(const MutationScore score) @trusted {
-        db.run(insert!MutationScoreHistoryTable, MutationScoreHistoryTable(0,
-                score.timeStamp, score.score.get));
+        auto stmt = db.prepare(format!"SELECT id FROM " ~ mutationScoreHistoryTable ~ "
```
joakim-brannstrom-work1, line 295: remove `format!`, it isn't needed

joakim-brannstrom-work1, line 295: i think this select is a left-over from implementing the new putMutationScore. it isn't needed anymore?

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -292,14 +292,43 @@ struct Database {
...
+                WHERE date('now') = date(time);");
+        auto res = stmt.get.execute;
+
+        if (!res.empty) {
+            auto id = res.front.peek!long(0);
+            auto updateStmt = db.prepare(
+                    format!"UPDATE " ~ mutationScoreHistoryTable
```
joakim-brannstrom-work1, line 302: missing binding the date

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -292,14 +292,43 @@ struct Database {
...
+            db.run(insert!MutationScoreHistoryTable,
+                    MutationScoreHistoryTable(0, score.timeStamp, score.score.get));
+        }
     }
 
     // Add a mutation score for the individual files
     void putFileScore(const FileScore score) @trusted {
-        db.run(insert!MutationFileScoreHistoryTable, MutationFileScoreHistoryTable(0,
-                score.timeStamp, score.score.get, score.file.toString));
+        auto stmt = db.prepare(format!"SELECT id FROM " ~ mutationFileScoreHistoryTable
```
joakim-brannstrom-work1, line 315: this is a left-over sql statement that can be removed?

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -292,14 +292,23 @@ struct Database {
 
     /// Add a mutation score to the history table.
     void putMutationScore(const MutationScore score) @trusted {
-        db.run(insert!MutationScoreHistoryTable, MutationScoreHistoryTable(0,
-                score.timeStamp, score.score.get));
+        auto stmt = db.prepare("INSERT OR REPLACE INTO " ~ mutationScoreHistoryTable ~ "
+                (score, time) VALUES (:score, :time);");
+        stmt.get.bind(":score", score.score.get);
+        stmt.get.bind(":time", score.timeStamp.toISOExtString().replace("T",
```
joakim-brannstrom, line 298: dont forget to call toUTC before stringification because otherwise the timeStamp is in local clock

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
```diff
@@ -292,14 +292,23 @@ struct Database {
...
     }
 
     // Add a mutation score for the individual files
     void putFileScore(const FileScore score) @trusted {
-        db.run(insert!MutationFileScoreHistoryTable, MutationFileScoreHistoryTable(0,
-                score.timeStamp, score.score.get, score.file.toString));
+        auto stmt = db.prepare("INSERT OR REPLACE INTO " ~ mutationFileScoreHistoryTable ~ "
+                (score, time_stamp, file_path) VALUES (:score, :time, :path);");
+        stmt.get.bind(":score", score.score.get);
+        stmt.get.bind(":time", score.timeStamp.toISOExtString().replace("T",
```
joakim-brannstrom, line 308: toUTC

plugin/mutate/source/dextool/plugin/mutate/backend/database/standalone.d
virrenn: I call it in test_mutant/package.d on line 1121, so it works with UTC time from the start

joakim-brannstrom: instead of calling replace you can call toSqliteDateTime in miniorm.

plugin/mutate/source/dextool/plugin/mutate/backend/test_mutant/package.d
```diff
@@ -1118,7 +1118,17 @@ nothrow:
         auto files = spinSql!(() => db.getFiles());
         const fileScores = reportScores(*db, kinds, files);
         const score = reportScore(*db, kinds);
-        const time = Clock.currTime;
+        auto time = Clock.currTime.toUTC;
+
+        try {
+            time.hour = 0;
```
joakim-brannstrom, line 1124: i think this is better to move to putMutationScore because then we "know" that it is always zeroed just before it is saved to the DB wherever it is called in the future.

PR #1620, opened 2022-08-15
mutate: testcase namefilter

OliverGreen27:
filter added for url and links involving testcase names. Generated text through js of testcase names have also been encoded to not create corrupted html files

### DIFF
--- a/plugin/mutate/source/dextool/plugin/mutate/backend/report/html/utility.d
+++ b/plugin/mutate/source/dextool/plugin/mutate/backend/report/html/utility.d
@@ -26,6 +26,23 @@ string pathToHtml(string p) {
     return p.pathSplitter.joiner("__").toUTF8;
 }
 
+string removeAllNonAlphaNum(string s) {
+    import std.ascii : isASCII;
+    import std.uni : byCodePoint, isAlphaNum;
+    import std.array : appender;
+    import std.conv : text;
+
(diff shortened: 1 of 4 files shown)


joakim-brannstrom:
This will probably work less well when multi-byte utf8 characters are involved. The loop iterates over code points which may break when e.g. chinese letters are used. This has been a bug before thus I think it is best to be careful.

See this function for how to iterate over grapheme https://github.com/joakim-brannstrom/dextool/blob/4a39750edee3942008b2ec685a3ee084953bb5a8/libs/dextool/source/dextool/compilation_db/package.d#L640

plugin/mutate/source/dextool/plugin/mutate/backend/report/html/utility.d
```diff
@@ -26,6 +26,22 @@ string pathToHtml(string p) {
...
 
+string removeAllNonAlphaNum(string s) {
+    import std.ascii : isASCII;
+    import std.uni : byCodePoint, isAlphaNum;
+    import std.array : appender;
+
+    auto result = appender!(char[])();
+    foreach (c; s.byCodePoint) {
+        if (isAlphaNum(c) && isASCII(c)) {
+            result.put(c.text);
```
joakim-brannstrom, line 37: need to import `std.conv : text` to make it compile

